PR #25, opened 2016-02-26
Improve error handling and tests for when API has problems

vitorbaptista:
We're returning HTTP code 502 Bad Gateway for when there were some problems
with the API. This seems to be the best error code, considering that this app
is basically a gateway to the OpenTrials API.

I've also refactored how we test for the responses. We have to do
`JSON.parse(JSON.stringify(data))` to convert JavaScript primitives (like Date)
into JSON. It's better to do that than to manually convert the dates to a
string before checking for the API response, as we were doing before, because
we would have to keep track of what fields need to be converted. Converting to
JSON allows us not to care about this.

### DIFF
--- a/handlers/trials.js
+++ b/handlers/trials.js
@@ -1,3 +1,4 @@
+const Boom = require('boom');
 const trials = require('../agents/trials');
 
 function trialsDetails(request, reply) {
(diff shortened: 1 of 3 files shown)


akariv:
looks good to me

PR #88, opened 2016-04-18
params validation has been added

borysyuk:


vitorbaptista:
@borysyuk The code looks good, just needing a few adjustments before merging.

Please, extract the validation to a separate method like `validateQueryParams(request.query)` and call it passing `query` on the first lines of `searchPage` like:

```javascript
const validatedQuery = validateQueryParams(request.query);
if (validatedQuery.error) {
  // handle the error and return
}
const query = validatedQuery.value;
```

Inside the `validateQueryParams`, you simply do `return Joi.validate(query, schema);`. Remember to use the query coming back from `Joi`, because it not only validates, but also casts the values to what we need (e.g. `page` as integer).

Please expand the tests as well. Ideally, they would fail if any of the schema validations is modified or removed. I understand that it might be difficult to do, specially because we need to mock HTTP calls, so I'm happy with a simple test per parameter (e.g. making sure `page` needs to be an integer, etc). It'll require some changes to the test suite, which I'm happy to help if needed.

handlers/search.js
```diff
@@ -111,6 +112,39 @@ function getFilters(query) {
...
+
+  const searchParams = {
+    page: query.page,
+    registration_date_start: query.registration_date_start,
+    registration_date_end: query.registration_date_end,
+    location: query.location,
+    q: query.q,
+    sample_size_start: query.sample_size_start,
+    sample_size_end: query.sample_size_end,
+  };
```
vitorbaptista: Why are you using `searchParams` instead of `request.query`? Ideally, we would use `request.query`, so we catch unrecognized query parameters as well.

handlers/search.js
```diff
@@ -111,6 +112,39 @@ function getFilters(query) {
 
 function searchPage(request, reply) {
   const query = request.query;
+
+  const schema = Joi.object().keys({
+    page: Joi.number().integer().min(1),
```
vitorbaptista: There's a maximum page as well: 100 (check https://github.com/opentrials/api/blob/master/api/swagger/swagger.yaml)

handlers/search.js
borysyuk: no problem, but i saw one test https://github.com/opentrials/opentrials/blob/master/test/handlers/search.js#L87
this test fails. i will use recognized param only.

vitorbaptista: Feel free to change the tests if they're wrong, just make sure the issue is with the test itself, and not the code.

borysyuk:
@vitorbaptista fixed.

handlers/search.js
```diff
@@ -64,7 +57,7 @@ function getPagination(url, currentPage, perPage, maxPages, totalCount) {
 function getFilters(query) {
   const filters = {};
   const ensureIsArrayAndQuoteElements = (values) => (
-    ensureIsArray(values).map((val) => `"${val}"`)
+    values.map((val) => `"${val}"`)
```
vitorbaptista: I haven't tested the code, but it seems that you're removing `ensureIsArray` because Joi converts single elements to array, right? If that's the case, the name of this function is now incorrect: it's just quoting elements, and not ensuring it's an array. Please, rename it accordingly (maybe `quoteElements()`))

handlers/search.js
```diff
@@ -103,36 +96,47 @@ function getFilters(query) {
...
+  return Joi.validate(query, schema);
+}
+
 function searchPage(request, reply) {
-  const query = request.query;
+  const validatedQuery = validateQueryParams(request.query);
+  if (validatedQuery.error) {
+    reply(
+      Boom.badRequest('invalid query')
+    );
```
vitorbaptista: Please, change this to `return reply(Boom.badRequest('Invalid query', validatedQuery.error)` (with the "Invalid" capitalized), so the rest of this function isn't executed.

vitorbaptista:
@borysyuk The code looks great! Adding `Joi` removed quite a bit of code that I wasn't expecting (like the `ensureIsArray`). Less code == better code :+1:

Regarding the styling changes (e.g. `{ foo: 'bar' } -> {foo: 'bar'}`). Did you change that because eslint was complaining? That would be surprising, considering it's currently passing without failures. Maybe some upgraded version?

borysyuk:
@vitorbaptista i use intellij IDEA, so when I re-format my code, IDE reformatted that block too.

borysyuk:
@vitorbaptista  fixed

vitorbaptista:
Great job, @borysyuk! I've squashed your commits and changed a few minor things on https://github.com/opentrials/opentrials/commit/e34a3626a94c15ec0334a366887dca2e8527aa31. This is now merged in master.

PR #90, opened 2016-04-22
Feature/instance pages

borysyuk:


test/handlers/search.js
```diff
@@ -81,6 +81,19 @@ describe('search handler', () => {
...
+        mockApiResponses();
+        return server.inject('/search?q=test&page=aa')
+          .then((_response) => {
+            _response.statusCode.should.equal(400)
+          });
+
+        return server.inject('/search')
+          .then((_response) => {
+            _response.statusCode.should.equal(502);
+          });
```
vitorbaptista: These lines will never be executed because there's a `return` before them.

routes/index.js
```diff
@@ -35,6 +35,28 @@ const routes = [
     method: 'GET',
     handler: require('../handlers/search'),
   },
+
+  {
+    path: '/problems/{id?}',
```
vitorbaptista: As the `id` is required, use `/problems/{id}` instead (same for the other routes)

views/interventions-details.html
```diff
@@ -0,0 +1,23 @@
+{% extends "layouts/two-columns.html" %}
+
+{% block bannerFooter %}
+<ul class="overview-list">
+  <li>
+      <span class="date">
+        Registration: {{ intervention.registration_date | formatDate }}
```
vitorbaptista: Where does this `registration_date` come from?

routes/index.js
borysyuk: i copied from this route path: '/trials/{id?}',
I think, id is required here too, right?

vitorbaptista: Yes, that was the remnants of a `/trials` route. I fixed it there as well in another pull request :+1:

vitorbaptista:
@borysyuk Could you fix the conflicts with master? Also, your IDE re-formatted the code again.

borysyuk:
@vitorbaptista I have merged and resolved conflicts

vitorbaptista:
Cool! I'll merge it as soon as https://github.com/opentrials/api/pull/36 is done, so we don't end up with an interface accessing an API that doesn't exist.

vitorbaptista:
Merged in https://github.com/opentrials/opentrials/commit/48ad147b5dd2073e93dd5b7cab4496bd7c708f43 :+1:

PR #91, opened 2016-04-25
Expose trialrecords

### LINKED ISSUE #58: On trial detail pages, add the section for register entries
# Description

The mocks for the trial detail pages have a section for registry entries of the trial [see here](http://opentrials.github.io/mockups/trial.html).

This section is important as it shows users the core information that we built our "ideal trial object" from. This section also should allow the user to see the record from the register as we collected it, and, follow a link to the original page or API call that returns that information.

# Tasks

* [ ] Implement the section of the page for registers
* [ ] Clicking on the element for each register should open a simple modal, which, in the simplest way possible, renders the data of the trial record from that register as a list of key: value
* [ ] Above the key: value list, have a template snippet like:
  - "This record was cached from {REGISTER_NAME} on {DATE_OF_ACQUISITION}. See the source data at this link: {LINK_TO_SOURCE_PAGE}"

vitorbaptista:
Fixes #58 

Depends on https://github.com/opentrials/api/pull/37


vitorbaptista:
@roll could you check this and https://github.com/opentrials/api/pull/37? The deployed instance at http://ot-explorer-next-pr-91.herokuapp.com is working, just check the `Sources` part in http://ot-explorer-next-pr-91.herokuapp.com/trials/05cc77ad-5575-4c04-9309-4c64d5144b07

roll:
ok will do soon

roll:
:+1: awesome! Our trial pages start to look usefull.

Issue #63, opened 2016-04-04
UI elements for users


pwalsh:
# Description

After https://github.com/opentrials/opentrials/issues/62 implement the UI for user auth flows. Like there, not that the [OS Packager](https://github.com/openspending/os-packager) and [OS Auth Client](https://github.com/openspending/os-auth-client) may have code of interest (or, not :)).

# Tasks

- [ ] UI/X for user flows
- [ ] Functional tests as relevant

PR #108, opened 2016-05-03
Feature/user login

### LINKED ISSUE #62: Allow users to authenticate and have roles (permissions)
# Description

Users need to be able to authenticate with the system, so that users can perform certain actions. Authenticated users need to be able to have roles, where each role relates to a flat permissions level (no granular permissions logic). Our use cases currently require three roles:

- `normal`
- `curator`
- `administrator`

No logic needs to be added on these roles at present, except that users will get a default `normal` role, and core developers can escalate a user's role to curator or developer by directly manipulating the database.

See the user stories related to [exploring data](https://github.com/opentrials/docs/blob/master/docs/extras/user-stories/index.md#a-exploring-the-database) and [curating data](https://github.com/opentrials/docs/blob/master/docs/extras/user-stories/index.md#b-contributing-and-curating-data) for some use cases related to users.

I ( @pwalsh ) do have a very specific idea of implementation in mind, based on many apps I've done with this implementation. We implement it in [OpenSpending](https://github.com/openspending/os-conductor/tree/master/conductor/blueprints/authentication), and it goes like this:

- Only oauth by existing providers (google is enough for now)
- All data comes from the provider, we essentially "cache" what we need (user name, etc) in a user table
- Client apps (like the explorer) get [JSON Web Tokens](https://jwt.io) with a payload of user identity info, and also info on expiring the token

Also note that the design by @smth for OT is based on the design in he new OpenSpending for users (in turn, based on the details of this implementation).

In other [OKI code](https://github.com/okfn/opendatacensus/blob/master/census/routes/utils.js#L22), we've used [Passport.js](http://passportjs.org) for user auth (albeit without JWT, and the full client-server interaction we use in OpenSpending and here)

# Tasks

- [ ] API methods to authenticate
- [ ] User gets a default role on first login
- [ ] Server returns JWT tokens for use by client apps
- [ ] Tests

vitorbaptista:
Fixes #62 and #63


vitorbaptista:
@roll Please review. Unfortunately, logging in doesn't work in the deployed PR instance on Heroku because of the way the OAuth credentials work (they require a specific domain to be authorized).

docker-compose.yml
```diff
@@ -1,22 +1,24 @@
 web:
```
roll: Do we use docker stuff somehow in this project?

test/common.js
```diff
@@ -1,9 +1,28 @@
...
+function clearDB() {
+  const tables = [
+    'oauth_credentials',
+    'users',
+  ];
+  let deferred = config.bookshelf.knex.migrate.latest();
+
+  for (const tableName of tables) {
+    // eslint-disable-next-line no-loop-func
+    deferred = deferred.then(() => config.bookshelf.knex(tableName).select().del());
```
roll: Could we call here something like `config.bookshelf.knex.migrate.<to null state>();` instead of manual table deletion? It will test our down migrations also.

vitorbaptista: Migration rollbacks are a bit tricky. It's not uncommon for migrations to be destructive ([example in OpenTrials API](https://github.com/opentrials/api/blob/master/migrations/20160415172210_remove_trials_trialrecords_table.js)), and so rolling back wouldn't work. We could drop the entire database and recreate it here, but I haven't found an easy way to do that with KnexJS (and it requires the user to be able to create databases).

docker-compose.yml
vitorbaptista: We don't. It might be useful to ease the project installation for contributors, but then we'd have to test it somehow. I'd prefer to keep it for now, but if we don't figure out an easy way for testing, remove it later :+1:

roll:
Migration fails because it can't find prev migrations files:

```
Error: The migration directory is corrupt, the following files are missing: 20160216120159_create_initial_schema.js, 20160407111633_create_trialrecords_entity_drop_records.js, 20160407111800_add_meta_fields_to_entities.js, 20160413110458_add_gender_and_has_published_results_to_trials_and_trialrecords.js, 20160415172210_remove_trials_trialrecords_table.js, 20160429190634_update_publications.js, 20160501171307_add_source_id.js, 20160501172805_remove_links_facts.js, 20160501175732_add_slug_and_facts.js, 20160502123658_add_diffs_table.js
``` 

It think the same will be when we will trigger this migration on production.

Anyway Is it OK to have two migrations sources for one database? Ideally I suppose better way to handle auth things on `api` level so `explorer` will be knowing nothing about database. Or we going to use here other database?

roll:
Aside this migrations/design problem it works very well - I could log in and log out with Google :+1:

vitorbaptista:
@roll Are you trying to use the same database for the API and the frontend? That won't work (as you've seen): you need a new database for the frontend. The reasoning behind this separation was to keep the API "dumb", responsible only for the trial data and search interfaces. We'll have quite a few functionalities around adding trial data, crowdsourcing of uploads, voting for deduplication, moderation, etc.. Doing that in the API level would make it much more complicated.

In this sense, the Explorer is enhancing the functionalities of the API.

roll:
Yea wrapping everything in `api` could be a mess.. Just never heard about our 3rd database) As I've written above - it works good with other database.

vitorbaptista:
Yeah, there're many changes in this PR. I need to write some diagrams and docs on the current architecture in the future, similar to what you've done with the collectors/processors.

PR #112, opened 2016-05-11
Feature/stats endpoint

borysyuk:


vitorbaptista:
@borysyuk instead of building the page using AJAX, that should be rendered in the server. Please create a new agent (see https://github.com/opentrials/opentrials/blob/master/agents/trials.js as an example) to get the data through the API and then pass it to the view. You can check the `trials` agent, [its handler](https://github.com/opentrials/opentrials/blob/master/handlers/trials.js) and [view](https://github.com/opentrials/opentrials/blob/master/views/trials-details.html) as examples.

Ideally, this PR shouldn't contain any client-side javascript.

views/stats.html
```diff
@@ -0,0 +1,57 @@
...
+    </ul>
+  </div>
+
+  <div class="stats-node">
+    <h3>Date when we updated each registry: </h3>
+    <ul class="meta" id="updatedRegistry">
+      {% for dateRegistry in stats.dateRegistry %}
+      <li>
+        <dt>{{ dateRegistry.name }}</dt>
+        <dd>{{ dateRegistry.updatedate | formatDate }}</dd>
```
vitorbaptista: This seems to be a residue from old code, as you've changed `updatedate` to `latest_update_date`

vitorbaptista:
Merged as https://github.com/opentrials/opentrials/commit/566a607377d12a245b4ce8f678fc3395301b1ee3

Issue #71, opened 2016-04-05
API to act on user-contributed data


pwalsh:
# Description

Users with either role curator or administrator need to be able to assess user-contributed data. The interface for this is implemented in https://github.com/opentrials/opentrials/issues/70 

# Tasks

- [ ] Enhance schema for data contribution records to support fields related to curator assessment
  - reject/accept; comments; assessing_user
- [ ] Add API action to edit data contribution records. This action is only accessible to `curators` and `administrators`
- [ ] Tests

vitorbaptista:
As per #62, this data is stored in the Explorer DB, and not in the API, so this issue won't be fixed. Please reopen if needed.

Issue #70, opened 2016-04-05
Interfaces to view and assess user-contributed data


pwalsh:
# Description

Requires https://github.com/opentrials/opentrials/issues/71

Users with either role `curator` or `administrator` need to be able to authenticate and access a view that shows all user contributed data. There is no existing design for this view, but that is ok. The main thing for this stage is to provide a simple way to allow these users to perform the following tasks:

- See data contributed manually by users
- Mark this data as accepted or rejected
- Add comments on the contribution

This set of tasks implements [user story B6](https://github.com/opentrials/docs/blob/master/docs/extras/user-stories/index.md#b-contributing-and-curating-data).

Note that this set of task does not try to handle what happens after data is accepted or rejected. We will have a manual process after this point for some time, and there is no need to write code for it at this stage.

# Tasks

- [x] When a user of with role `curator` or `administrator` is logged in, show an additional top-level menu item: "New data"
- [x] Create a new view for "new data" that is only accessible to authenticated users with role `curator` or `administrator`
- [x] On the "new data" view, show a list of contributed data (quantity of contributed data will be small for quite some time, so put no effort into search/filter etc on the list, just order by the timestamp of when the data was added)
  - User name
  - Date
  - Link to record for the contributed data
  -  Indicator if the contribution has been reviewed or not
- [x] The detail view for a record of contributed data is also only accessible to `curators` and `administrators`
- [x] The detail view for a contributed data record has the following elements
  - [x] User name
  - [x] Date
  - [x] Associated Trial ID (link from it to the actual trial detail page)
  - [x] Link(s) to contributed data sources (so this user can download these sources for inspection)
  - [x] A form to enter assessment details
    - [x] Boolean to "Reject" or "Accept" the data
    - [x] "Comments" text area to make comments on the data contribution
    - [x] Action to submit form

Issue #68, opened 2016-04-05
Storage API for user-contributed files


pwalsh:
# Description

Any user who has authenticated can add data. This data may be associated with a known trial, or not. Such data never directly enters the main database. Rather, it goes to a staging area in the Warehouse so that staff can look at it and assess. The "look at it and assess" part is currently a manual process (and not in scope for this issue).

- Some user facing actions to add files can be see on http://opentrials.github.io/mockups/trial.html, to provide context for how this feature is presented to users

# Tasks

- [ ] Add a manual data input processing structure to the warehouse
  - Path in S3 bucket as a "staging area" for user-contributed data
  - Table for metadata on user contributed data
    - user ID
    - file path(s) on S3
    - notes per file (maybe file paths and file notes in a JSONB field for flexibility)
    - timestamps, etc.
    - related trial - the ID of a trial, IF the data relates to a trial
- [ ] Expose API endpoint to accept such data/meta data from any authenticated user
  - Note we have code in OpenSpending related to auth/z for file upload to S3. @roll contributed to some of that code.
  - This endpoint probably needs to be on the public API (and not some new API that is warehouse-specific), so this is the first public API that directly talks to the warehouse database
    - Unless we do what a warehouse-specific API that integrates with auth? Can discuss.
- [ ] Tests

vitorbaptista:
As per our discussion in #62, users are saved in the Explorer DB, leaving the API to be mostly (all?) read-only, responsible only for searching and exposing our trials and related entities. This means that this issue won't be fixed, as the data upload process work in the Explorer side. See https://github.com/opentrials/opentrials/pull/114 for more details and reopen if needed.

PR #114, opened 2016-05-13
Feature/contribute data

### LINKED ISSUE #69: Interfaces to add data
# Description

Requires https://github.com/opentrials/opentrials/issues/68

Users can add data to the system. This data can be associated with a known trial, or not. See here for data structure info and more general context https://github.com/opentrials/opentrials/issues/68

# Tasks

- [x] Implement interactions on trial detail pages
- [x] Add a new page "Contribute Data" that implements the upload data form, without the context of a trial (don't worry if design is not perfect)
- [x] Above two tasks require success/fail interactions in front of user:
  - [x] Can add if not authed
  - [x] Feedback if file upload failed
  - [x] Feedback if file upload successful
- [x] Add a link to the "Contribute Data" page from the footer
- [x] functional tests as relevant

vitorbaptista:
Add `/contribute-data` page where users (logged in or not) can upload documents related to a trial or not. The data is uploaded to S3 and its metadata saved to the database. Currently, there's no interface to see what was uploaded. In #70 and #71, we'll build a moderation interface for curators to approve/reject the files.

Fixes #69. Issue #68 won't be fixed, as the data goes to the Explorer database, not the API's.


pwalsh:
@vitorbaptista shouldn't metadata go straight to the warehouse, for (potential processing there)?

vitorbaptista:
@pwalsh I added it in the Explorer database because the users/roles are already in this DB and #70 would be created on top of it as well.

If we were to send the data contributions directly to the warehouse, then the moderation should be done over there as well IMHO, possibly as a separate codebase.

pwalsh:
@vitorbaptista no worries. i guess we can decide at a later date if there is a need for it.

roll:
For `processors` there is no difference from which database get data to process. But if `explorer` write data to `warehouse` any `warehouse` migration or outage could break `explorer`.

roll:
@vitorbaptista 
I'll review on Monday. Feel free to merge if it will be blocking - I'll ccould heck it merged.

roll:
@vitorbaptista 
It looks like I have problem with s3 settings. Trying to fix it. 

But error message not really descriptive:
- bad upload finishes with `{"statusCode":400,"error":"Bad Request","message":"child \"response\" fails because [\"response\" is not allowed to be empty]","validation":{"source":"payload","keys":["response"]}}` in a browser
- and `
160517/164338.618, [response], http://localhost:5000: post /contribute-data {"trial_id":"2b1c993b-44bd-42d7-b399-9649d8974a1a","redirectTo":"/trials/2b1c993b-44bd-42d7-b399-9649d8974a1a"} 400 (20ms)` in logs

roll:
@vitorbaptista 
So a scenario:
- go to http://localhost:5000/contribute-data
- select and upload file with size lesser than
- get `{"statusCode":400,"error":"Bad Request","message":"child \"response\" fails because [\"response\" is not allowed to be empty]","validation":{"source":"payload","keys":["response"]}}` and no logged error in console

I think problem with my setup but I suppose we need to add some error logging here.

PR #141, opened 2016-05-18
Feature/publication endpoint

borysyuk:


views/publications-details.html
```diff
@@ -0,0 +1,40 @@
+{% extends "layouts/two-columns.html" %}
+
+{% block main %}
+<article>
+  This publication was last updated on
+  <time datetime={{record.updated_at}}>{{ publication.updated_at | formatDate }}</time>
```
vitorbaptista: Shouldn't it be `{{ publication.updated_at }}` instead of `record`?

views/publications-details.html
```diff
@@ -0,0 +1,40 @@
...
+
+  <h2>Facts</h2>
+  <ul class="overview-list">
+    {% for fact in publication.facts %}
+    <li>{{ fact }}</li>
+    {% endfor %}
+  </ul>
+
+  <h2>Slug</h2>
+  {{publication.slug}}
```
vitorbaptista: No need for a slug here, as it's only used to generate URLs

test/handlers/publications.js
```diff
@@ -0,0 +1,60 @@
...
+    describe('API is OK', () => {
+      const publication = JSON.parse(JSON.stringify(
+        fixtures.getPublication()
+      ));
+      let response;
+
+      before(() => {
+        apiServer.get('/publications/'+publication.id).reply(200, publication);
+
+        return server.inject('/publications/'+publication.id)
```
vitorbaptista: For generating these strings, prefer ES6 style:

```
apiServer.get(`/publications/${publication.id}`)
```

test/fixtures/publications.js
```diff
@@ -0,0 +1,20 @@
...
+    },
+    source_url: 'https://clinicaltrials.gov/ct2/show/NCT00000000',
+    created_at: new Date('2016-01-01'),
+    updated_at: new Date('2016-02-01'),
+  };
+
+  return problem;
+}
+
+module.exports = getProblem;
```
vitorbaptista: Rename `problem` to `publication` over this file

test/handlers/publications.js
pwalsh: big 👍 on ES6 style string templates. we should use them everywhere.

test/agents/publications.js
```diff
@@ -0,0 +1,23 @@
+'use strict';
```
vitorbaptista: Always add a newline after `'use strict;'`

views/trials-details.html
```diff
@@ -78,5 +78,23 @@
...
+    {% for publication in trial.publications %}
+      <dt>
+        <a href="/publications/{{ publication.attributes.id }}">
+          {{ publication.attributes.title }}
+       </a>
+      </dt>
+      <dd>
+        Last updated on
+        <time datetime="{{ publication.attributes.updated_at }}">
+          {{ publication.attributes.updated_at | formatDate }}
```
vitorbaptista: No need for the last updated date. Just show the publication title here, probably using a `<ul>` instead of a `<dl>`

views/publications-details.html
vitorbaptista: Actually, just remove this last updated time here. I don't think we keep updating publications as we do with sources, so there's no need for it.

test/fixtures/publications.js
```diff
@@ -0,0 +1,20 @@
...
+      id: uuid.v1(),
+      name: 'nct',
+      type: 'register',
+    },
+    source_url: 'https://clinicaltrials.gov/ct2/show/NCT00000000',
+    created_at: new Date('2016-01-01'),
+    updated_at: new Date('2016-02-01'),
+  };
+
+  return problem;
```
vitorbaptista: There are still some `problem` instances here.

views/trials-details.html
```diff
@@ -78,5 +78,17 @@
       </dd>
     {% endfor %}
   </dl>
+
+  <h2>Publications</h2>
+  <ul>
+    {% for publication in trial.publications %}
+      <dt>
```
vitorbaptista: This should be an `<li>` , not `<dt>`

views/publications-details.html
```diff
@@ -0,0 +1,34 @@
...
+
+{% block main %}
+<article>
+  <h2>Publication description</h2>
+  <ul class="overview-list">
+    <li>{{ publication.title }}</li>
+  </ul>
+
+  <h2>Source</h2>
+  <a href="{{publication.source_url}}">{{publication.source.name}}</a>
```
vitorbaptista: Sorry, I haven't ran the code so I can't answer it. Where does `publication.source_url` points to?

borysyuk: I don't know. I have almost empty DB and there was clear table, but this table has `source_url` field.
If we don't need to display this info, I can remove that link.

views/trials-details.html
borysyuk: @vitorbaptista 
We have many places with <dl> and <dd> in this view. Why we should use <ul> and <li> in this case only?  Should i change all <dl> and <dd> in this view?

vitorbaptista: So, `<dl>` are definition lists. They need a term (`<dt>`) and its value (`<dd>`). We're now just showing a list of publications, we would have only the terms (`<dt>`), which isn't valid in HTML5 (https://www.w3.org/TR/html-markup/dl.html)

This is different from the others, because they all have terms and values. To be honest, I'm probably abusing the `<dl>` element in some parts here (e.g. "Sources") so I might end up changing it to regular `<ul>` lists later. For now, just change publications for `<ul>` and `<li>` :+1:

views/publications-details.html
vitorbaptista: No worries, we can see later and change if needed.

roll: `source_url` is the same as for other entities - where it was collected/scraped/downloaded

test/fixtures/api/swagger.yaml
```diff
@@ -525,6 +555,54 @@ definitions:
...
+        type: string
+      title:
+        type: string
+      abstract:
+        type: string
+      authors:
+        type: array
+        items: {
+          "type": "string"
+        }
```
vitorbaptista: Same issue here with, instead of `items: { "type": "string" }`, use `items: type: string` (with newlines)

vitorbaptista:
@borysyuk Could you please rebase from master and get the new `swagger.yaml` from the API's master? There were some changes there what would break this code, mainly now we don't have `publications.attributes` anymore.

borysyuk:
@vitorbaptista I don't like to do `git rebase` because it is potentially dangerous operation. Can I make PR?

views/trials-details.html
```diff
@@ -178,5 +178,17 @@ <h2 class="disabled">{{ name }}</h2>
       </dd>
     {% endfor %}
   </dl>
+
+  <h2>Publications</h2>
+  <ul>
+    {% for publication in trial.publications %}
+      <li>
+        <a href="/publications/{{ publication.attributes.id }}">
+          {{ publication.attributes.title }}
```
vitorbaptista: This won't work after the last rebase because of changes in the API response

vitorbaptista:
Merged :+1:

Issue #122, opened 2016-05-16
Fix main navigation


pwalsh:
# Description

Requires https://github.com/opentrials/opentrials/issues/121

# Tasks

- [ ] Search: rename to "Explore" and remove the icon
- [ ] Login: inline with other nav items
- [ ] Order of nav items:` Explore | About | Login [OKI PANEL]`

<img width="330" alt="screenshot 2016-05-16 19 21 22" src="https://cloud.githubusercontent.com/assets/791678/15296011/082de086-1b9c-11e6-82dc-41ab96f2d542.png">

pwalsh:
@smth my suggestion for inlining login is just because I'm not sure how it will work with the OKI Panel. However, you will have a solution for this, so, go ahead with any solution you prefer of course.

Issue #121, opened 2016-05-16
Integrate OKI Panel Bar and Cookie Bar


pwalsh:
# Description

As with most OKI properties, integrate the Panel and the Cookie Bar.

# Tasks

- [x] Add OKI Panel ( http://a.okfn.org/html/oki/panel/ )
  - [x] Shift the existing menu to the left some amount so the panel does not cover it
- [x] Add Cookie Bar ( http://a.okfn.org/html/oki/cookie-bar/ )
  - [x] Only show cookie bar if google analytics is set

smth:
- OKI Panel is added, with slight menu layout changes to accommodate it. 
- The cookie bar is added in [this branch](https://github.com/opentrials/opentrials/tree/cookie-bar), but currently is not dependant on Google Analytics (there is no Google Analytics). Not sure if should be merged as is.

pwalsh:
DONE.

PR #168, opened 2016-05-23
Ui improvements

smth:


vitorbaptista:
@smth is this good to go?

smth:
Yup.

vitorbaptista:
@smth The changes look good, there're just a few comments

* Why changing "Search" to "Explore"? I think "Search", being more common, has a clearer meaning;
* There must be a "Contribute Data" (or similar) link, so an anonymous user can add data not necessarily related to a trial. Maybe we could add a text like "Is this related to a trial? If so, please search for the trial and contribute data from there, or give us the trial ID here" (with a better wording), just in case the data is related to a trial;
* Do we really need the OK Ribbon? It looks alien to the rest of the page, and is too prominent for something that probably most of our users won't be interested in (not to mention the extra KBs added)

smth:
I'm just carrying out tasks (https://github.com/opentrials/opentrials/issues/121, https://github.com/opentrials/opentrials/issues/122) from @pwalsh at this point. It sounds like the "Contribute Data" link maybe should have remained though?

My 2¢ on the ribbon thing is, I don't like it (anywhere). I feel this is a bigger conversation that needs to be had though.

vitorbaptista:
@smth OK, so from my comments just re-add the "Contribute Data" link. Part of the requirements for contributing data was that it doesn't need to be related to a trial, so that's still needed.

We'll talk about the ribbon offline :wink:

smth:
[Done](https://github.com/opentrials/opentrials/pull/168/commits/bcc2a37fdbd3713582f5c5ead06505df82893c35)

views/partials/main-menu.html
```diff
@@ -4,6 +4,11 @@
   </a>
 </li>
 <li>
+  <a href="/search">
+    Contribute Data
```
vitorbaptista: That should be `/contribute-data`

smth: Updated.

vitorbaptista:
Merged :+1:

PR #169, opened 2016-05-23
Feature/search results for entities

borysyuk:


assets/styles/_search-results.scss
```diff
@@ -1,4 +1,4 @@
-.result {
+.result, .result-small {
```
vitorbaptista: each selector should be in a single line

assets/styles/_search-results.scss
```diff
@@ -59,6 +59,13 @@
   }
 }
 
+.result-small {
+  padding-bottom: 0;
+
+  h2.title  {
+    font-size: em(14);
+  }
+}
```
vitorbaptista: add newline

agents/trials.js
```diff
@@ -57,8 +57,13 @@ function searchTrials(query, page, perPage, filters) {
     .then((response) => response.obj);
 }
 
+function searchTrialsByEntity(entityType, entityName) {
+  return searchTrials(`${entityType}:${entityName}`, 1, 10);
```
vitorbaptista: This will break if the `entityName` has whitespaces. You must escape `entityName` (make sure there're no unescaped `"` in it) and use `${entityType}:"${entityName}"`. This isn't enough to be sure it handles all cases, but should do for now. I've created a new issue to do it later: https://github.com/opentrials/opentrials/issues/172

vitorbaptista:
@borysyuk This all looks good for @smth to work with, but I wonder why you only added the view to `persons-details.html`.

This should be OK, as this is just part of what's needed for @smth to work, but just wondering if there was some other reason.

agents/trials.js
borysyuk: @vitorbaptista: I think, we don't should escape `entityName` at this side. I have tested request `person:hello " world` and i got escaped value on API side `person:hello \" world` (and yes, i got some break)
In my opinion, we should process query string on API side only.

vitorbaptista: That would be ideal, but unfortunately isn't possible. For example, say that the API is called like http://api.opentrials.net/v1/search?q=location:united%20kingdom, we can't know from the API if the original query should've been `location:"united kingdom"` or `location:united kingdom` (i.e. `kingdom location:united`). The only place that knows what the query should be is the Explorer, so this logic needs to be there.

PR #183, opened 2016-05-27
Feature/google analytics

borysyuk:


config/nunjuck.js
```diff
@@ -1,6 +1,8 @@
 'use strict';
 
 const path = require('path');
+require('dotenv').config();
```
vitorbaptista: Add the `require('dotenv').config()` before the `const path = require('path')`. This won't change much here, but it's good to keep the pattern of, when needed, running `dotenv` before anything else.

.env.example
```diff
@@ -15,3 +15,4 @@ S3_BUCKET=s3-bucket
 S3_REGION=s3-region
 S3_CUSTOM_DOMAIN=https://s3-custom-domain-if-it-exists.com
 MAX_UPLOAD_SIZE=size-in-bytes
+GOOGLE_ANALYTICS_CODE=
```
vitorbaptista: Just to keep it consistent with the rest of the file, could you change it to something like `GOOGLE_ANALYTICS_CODE=google-analytics-code`

vitorbaptista:
All good to go, just these minor adjustments :+1:

borysyuk:
@vitorbaptista done.

vitorbaptista:
Merged :+1:

Issue #50, opened 2016-04-04
Diff reports across threaded trial records


roll:
@pwalsh 

# Description

Clinical trial data is published in a range of places, including over multiple trial registers. It is a known issue that different publications can declare discrepant facts on a trial. When we have a set of threaded records, we want to run a "diffing" algorithm over the records and pull out these discrepancies as data points in their own right.

# Tasks

* [x] Write a diff algorithm for a set of threaded records
* [x] Store the results of such diffs. They are a property of the "ideal" model of the trial on which the records are matched
* [ ] Push this data forward from the warehouse to the main database/API

pwalsh:
Hey @roll 

Ben gave me an example query that can be useful for him and some researchers. I think maybe you can run a query like this as part of this work when you start it, and just export the actual results (query statement + results) and share with Ben.

He asks:

* get results of a query which selects trials that appear in 1+ registers, that have discrepancies on study end date, and one of those is "no end date", meaning, no end date is recorded in that particular record. We'd want that query result. Further, it would be good to filter those results, based on results where match is on trial ID, and results with match based on scientific title

roll:
@vitorbaptista 
I've rebased both PRs - so it's ready to merge. Please re-assign me if/when (to start processor for production after migration) other actions from me will be needed.

vitorbaptista:
@pwalsh I extracted the report you described on https://github.com/opentrials/opentrials/issues/50#issuecomment-206537109 to #197, as we're closing this. It might be good to tag those requests in a specific label :+1:

PR #184, opened 2016-05-27
[#50] Add Trials' discrepancies page

vitorbaptista:
@roll This is the Explorer-side of the discrepancies. It basically generates the discrepancies dynamically in `agents/trials.js#getDiscrepancies`, following a structure similar to what you've done for the `discrepancies` table.

Again, as this is dynamically generated, we might have a performance hit. I'm not concerned with it, though, because I think it would be easy to add caching later :+1:

Thoughts?

### DIFF
--- a/agents/trials.js
+++ b/agents/trials.js
@@ -1,5 +1,6 @@
 'use strict';
 
+const _ = require('lodash');
 const opentrialsApi = require('../config').opentrialsApi;
 
 function getTrial(trialId) {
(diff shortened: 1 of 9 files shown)


vitorbaptista:
@roll The PR deployments of both this PR and the API's is working. You can see the discrepancies page in https://ot-explorer-next-pr-184.herokuapp.com/trials/05cc77ad-5575-4c04-9309-4c64d5144b07/discrepancies

roll:
awesome :+1: 

also we need to add search field for `has_discrepancy`?

vitorbaptista:
@roll Great! Merged for now. We'll add a search option for filtering just trials with discrepancies later :+1:

PR #190, opened 2016-05-30
Adds OKI cookie bar

pwalsh:
This PR was mostly done by @smth 

I just integrated the if statement for the google analytics code


views/partials/scripts.html
```diff
@@ -13,5 +13,13 @@
 <script src="{{ 'js/commons.min.js' | revPath }}" defer></script>
 <script src="{{ 'js/vendor.min.js' | revPath }}" defer></script>
 <script src="{{ 'js/main.min.js' | revPath }}" defer></script>
-<script src="//a.okfn.org/html/oki/panel/assets/js/frontend.js" type="text/javascript" defer></script>
+<script src="http://assets.dev/a/html/oki/panel/assets/js/frontend.js" type="text/javascript" defer></script>
```
vitorbaptista: Can't access `http://assets.dev`. Also, if the JS is available over HTTPS, I'd rather have protocol-agnostic links for whenever we enable SSL in OpenTrials :+1:

smth: That change is incorrect, and should be ignored.

vitorbaptista:
Merged

PR #201, opened 2016-06-06
[#170] Escape queries to ElasticSearch

### LINKED ISSUE #170: Escape queries to ElasticSearch
The query string has a few reserved characters (see https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html#_reserved_characters). We need to escape them before passing strings to ElasticSearch.

For a way to do it, check https://github.com/elastic/elasticsearch-js/issues/257

As a longer-term solution, we might rethink our use of query strings, or investigate if [simple query strings](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-simple-query-string-query.html) help us.

vitorbaptista:
As we're using ES' query strings, we need to escape the values we send through
to make sure we don't end up sending some invalid queries (like "foo("). Most
of the sanitization code comes from
https://github.com/lanetix/node-elasticsearch-sanitize/. I couldn't use their
package directly because they also escape whitespaces and other things (e.g.
OR, AND, NOT), which we don't want. The code is trivial, though, so no big
deal.

Fixes #170


roll:
:+1:

Issue #189, opened 2016-05-30
Should be able to link data contributions to trials (and other entities)


vitorbaptista:
For example, if a person clicks on "Contribute data" in a specific condition, we should save that relationship. In other words, that data contribution should have a relationship to the condition.

pwalsh:
@vitorbaptista definitely. That matches one of the user stories of course :) https://trello.com/c/k6wtkFGH/19-as-a-contributing-user-i-want-to-be-able-to-add-data-to-opentrials-related-to-a-given-trial-from-a-given-trial-page-so-i-can-enr

vitorbaptista:
@pwalsh This is a bit bigger than that. It's currently possible to contribute data related to a trial, but we have links to contribute data in other places. For example, http://explorer.opentrials.net/conditions/f79f1404-529c-45e0-accd-2f49ce7e7d71. So this issue is about being able to link data contributions to any entity, and not only trials.

vitorbaptista:
We don't have entity pages anymore, so this isn't relevant.

PR #213, opened 2016-06-08
[#198] Add data categories and allow links to be contributed

### LINKED ISSUE #198: Add categories and allow links to be posted in data contributions
The categories should be:

* Article
* Clinical study report
* Analytics code
* Trial paperwork
  * Patient information sheet
  * Consent forms
  * Case report forms
  * Protocols
  * Statistical analysis plan
  * Lay summaries
* European Public Assessment Report (EPAR) or FDA document
* Other

For links, they could be:
* Registry entry
* Press release
* Trial webpage
* Critical review
* Blog post
* Other

Considering that changing the options depending on the contribution type (if upload or link) is a bit more complicated, just use a single list for everything.

vitorbaptista:
The link allow users to contribute for example blog posts or trials webpages.
Ideally, when a user sends a contribution, she'll either upload a file or add a
link (or both). However, I've chosen not to require this (for now, at least)
because of the added complexity to the form. It's easy to ignore/remove bogus
contributions as well, and we reduce friction by being more lenient with what
we receive.

For the categories, I've chosen to create another entity and save those to the
database so it's easier to tweak names and add translations in the future. This
has the drawback of the added complexity, a small performance hit, and making
it harder for us to accept free-form categories if needed later. However, I
thought it was worth the hassle.

Fixes #198


roll:
@vitorbaptista 
It looks like there is no any indication in logs when s3 upload fails

roll:
Can't check s3 upload because again something wrong with my bucket (and I can't debug it without logs) but categories work fine. A few thoughts:
- on trial page we have `contribute` link for persons, locations etc but clicking on it we get contributing form I would say not really obvious for user in this case. For example for location main field could be a text area but now it's comment (feels like additional element).
- the same for categories - ti could contain locations, persons etc also

:+1:

vitorbaptista:
Hey @roll

> Can't check s3 upload because again something wrong with my bucket (and I can't debug it without logs)

I need to improve the error handling/logging. I'll do it as part of a broader issue to send logs to Papertrail as well.

> * on trial page we have contribute link for persons, locations etc but clicking on it we get contributing form I would say not really obvious for user in this case. For example for location main field could be a text area but now it's comment (feels like additional element).

Good point! I'm avoiding to spend too much time on this for now, though, creating multiple forms before knowing if people will start uploading data or not. This is something to keep in mind when analyzing what people are doing (and during our user testings) :+1:

> * the same for categories - ti could contain locations, persons etc also

Added categories for conditions and treatments for now. We can add more when we start seeing what people send us

> when we on condition/intervention/publication page we can contribute data bit it's not linked to this object (for example except trial_id table data_contributions could have entity + entity_id fields to accomplish this goal)

Agreed. I've created an issue for this on #189

roll:
:+1: for PR

About multiple forms - may we don't need multiple form at all - I mean we could just improve UX (cc @smth) of current from like having multi values [accordion](https://jqueryui.com/accordion/) where use could choose types of data he adding (text, url, file). For different entities we could have predifined configurations for this form.

vitorbaptista:
@roll I agree that when you go to the contribute data page, there're no indication on what's it about, and which entity I'm adding data to. The user can feel lost. I think this issue will be solved in part when we add breadcrumbs (#216).

Also, there's no indication that the user isn't required to upload files: she can add only a link. The name of the page itself (contribute data) is written towards uploading files (#217).

Anyway, there're plenty of improvements to be made, and I'm sure we'll find many others after we start showing it to people. The most important for us to do now is to keep track of it, test, and monitor where users are getting stuck.

Thanks for the thorough review. It was very helpful! :+1:

PR #214, opened 2016-06-09
[#203] Fix UI bug when a trial had no registration_date

### LINKED ISSUE #203: Trials without registration_date are displayed as if the registration date is today
For example, http://explorer.opentrials.net/trials/591f65e3-385c-4734-8822-9a4323b3baa7. This trial has no registration date (could be confirmed by checking its JSON at http://api.opentrials.net/v1/trials/591f65e3-385c-4734-8822-9a4323b3baa7), but it's displayed as if it had.

The issue is that the `moment` library, when called with `undefined`, returns the current date. We should avoid this behavior by checking for undefined in `views/filters/format-date.js`, and also check that there is a registration date wherever it's used.

vitorbaptista:
The issue is that the `moment` library, when called with `undefined`, returns
the current date. We avoid this behavior by checking for `undefined` in
`views/filters/format-date.js`. I've also changed the UI to display a message
when we have no registration date.

Fixes #203


roll:
:+1:

Issue #209, opened 2016-06-07
The "study results" filter retuns the same trials for "all" and "trials without results" and isn't persistent across page reloads


vitorbaptista:
If you go to http://explorer.opentrials.net/search?has_published_results=true, note that the `Study results` select option is "All", and not "Trials with published results" as expected. The issue here is that Joi is converting `has_published_results=true` to a boolean, and we're checking for it as a string in `views/partials/search-form-fields.html`

vitorbaptista:
If `has_published_results == false`, we weren't adding any filters, so we would
get all trials (including those with published results). This is because we
relied on simply testing for truthyness, and not for the actual true/false
value. As `undefined` is falsy in JavaScript, there was no difference between
setting `has_published_results = false` and not setting `has_published_results`
at all.

vitorbaptista:
Fixed

PR #215, opened 2016-06-09
[#209] Fix bugs with published results filter

vitorbaptista:
If `has_published_results == false`, we weren't adding any filters, so we would
get all trials (including those with published results). This is because we
relied on simply testing for truthyness, and not for the actual true/false
value. As `undefined` is falsy in JavaScript, there was no difference between
setting `has_published_results = false` and not setting `has_published_results`
at all.

I've also fixed the bug where the study results filter wasn't persistent. The
problem here is that we were testing for "true" and "false" strings, and not
boolean values. This was the case before we added Joi for validation. As Joi
also converts the values, it stopped working.


roll:
:+1:

PR #218, opened 2016-06-09
Fix flashes of unstyled parts of the page

vitorbaptista:
This fixes both the flash of the mobile menu and the one of the advanced search. The advanced search has a drawback: after this patch, it won't be possible for users with JS disabled to use the advanced search.

I think nowadays we can safely assume that the vast majority of our users will have JS enabled, so it's OK. I have coded a workaround to allow non-JS users to also access the advanced search but, after thinking a bit more, I thought it wasn't necessary.

@smth We were using Modernizr here to only hide the advanced search fields when JS is enabled. Do you remember if we're using it in other places as well, or we can remove it?

### DIFF
--- a/assets/js/index.js
+++ b/assets/js/index.js
@@ -52,7 +52,8 @@ $(document).ready(function() {
   });
 
   // home page
-  $(".toggle-advanced").click(function() {
+  $(".toggle-advanced").click(function(e) {
+    e.preventDefault();
     $(".advanced").slideToggle("slow", function() {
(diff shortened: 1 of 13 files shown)


roll:
AFIAK it looks good

smth:
I don't think we're using it elsewhere.

vitorbaptista:
Great! Removed Modernizr and ended up adding the code to allow the advanced search to be used even without JS.

PR #219, opened 2016-06-10
[#198] Fix bug with file uploads and improve e2e tests

vitorbaptista:
We have to add all extra form fields to S3's policy, otherwise S3 will complain
that it received an unknown value.


roll:
:+1:

PR #232, opened 2016-06-17
[#231] Add filter for publications

### LINKED ISSUE #231: Add filter for trials with/without publications
Similar to the filter for discrepancies

vitorbaptista:
So the user is able to look for trials with and without publications.

Fixes #231


roll:
:+1:

PR #233, opened 2016-06-20
[#227] Escape Query String to avoid syntax errors

### LINKED ISSUE #227: Forward slash / in search box gives '502 Bad Gateway' error
I was searching for:

Hyaluronan and N-ERC/Mesothelin as Key Biomarkers in a Specific Two-Step Model to Predict Pleural Malignant Mesothelioma

http://explorer.opentrials.net/search?q=Hyaluronan+and+N-ERC%2FMesothelin+as+Key+Biomarkers+in+a+Specific+Two-Step+Model+to+Predict+Pleural+Malignant+Mesothelioma&sample_size_start=&sample_size_end=&gender=&has_published_results=&has_discrepancies=&registration_date_start=&registration_date_end=

And got the above error message - removing the / stops the problem.

vitorbaptista:
We're exposing ElasticSearch's query strings in the API to allow advanced
search features without relying on explicit fields. This has the drawback of
making it possible for the user to do an invalid query, which would raise an
syntax error. Obviously, this is bad UX.

ElasticSearch has a simple query string which avoids this, but we're using a
few query string features that aren't available in simple query strings like
searching for date ranges and in specific fields.

In the future, we will have to stop using query strings and build our own
syntax (similar to GitHub or Slack's search). Meanwhile, this change escapes
the query string passed by the user, so it's impossible to raise syntax errors.
This has the drawback of limiting what the user can do in the query. After this
change, she'll only be able to do simple queries, at most with logical
connectors like AND, OR and NOT.

This is a reasonable tradeoff for now, IMHO.

Fixes #227


roll:
:+1:

PR #239, opened 2016-06-22
Feature/under construction disclaimer

vitorbaptista:
This PR also refactors the view layouts, removing duplications. When merging, please don't squash the commits.


roll:
:+1: 

In addition to `in development` not we could also add a permanent note about nature of our data like it's there is always possibility of incorect data etc. Because we will never be having 100% clean data.

vitorbaptista:
@roll Agreed. We will have a static page talking about the data, how we extract stuff, etc etc...

PR #255, opened 2016-07-11
[#251] Filter to display numbers using en-GB locale

### LINKED ISSUE #251: Style numbers in explorer
For example, instead of 331504, show 331,504. To fix this issue, simply create a Nunjucks filter in https://github.com/opentrials/opentrials/tree/master/views/filters and use it everywhere we display numbers. For a similar example, check the https://github.com/opentrials/opentrials/blob/master/views/filters/format-date.js filter.

Use UK's number pattern (i.e. 1234.56 should be displayed as 1,234.56)

nightsh:
* Fixes #251

### DIFF
--- a/test/views/filters/format-number.js
+++ b/test/views/filters/format-number.js
@@ -1,0 +1,15 @@
+'use strict';
+
+const should = require('should');
+const formatNumber = require('../../../views/filters/format-number');
+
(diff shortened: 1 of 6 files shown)


nightsh:
test coming soon, do not merge yet :)

PR #256, opened 2016-07-11
Add bourbon dependency, mention gulp requirement in README

nightsh:


README.md
```diff
@@ -14,6 +14,7 @@ OpenTrials is an app to explore, discover, and submit information on clinical tr
 ### Requirements
 
 * Node 5.8
+  * gulp
```
vitorbaptista: As `gulp` is in our `package.json`'s dependencies, this shouldn't be needed. You could argue that `gulp` shouldn't be in our dependencies, but we need it there for building on Heroku.

nightsh: Then I'll just drop it, hang on

Issue #93, opened 2016-04-27
Implement custom error pages


vitorbaptista:
Currently, the user sees a JSON response (e.g. http://explorer.opentrials.net/search?page=0). This might help: http://stackoverflow.com/questions/29571703/how-can-i-customize-the-validation-error-response-in-hapi-js

This should be implemented for both HTTP 4xx and 5xx errors

nightsh:
As I cannot assign it to myself, I'll comment here to let everyone know I'm on it now, will be back with either some questions or a PR ;)

vitorbaptista:
Fixed in #261

PR #261, opened 2016-07-12
[#93] Add error pages

nightsh:


vitorbaptista:
@nightsh The code looks good. I was surprised on how simple it was (thanks, hapi! :smile: )

The error page design could be improved, though. It currently looks like:

404 error

I don't want to spend too much time on the UI, but there're some low-hanging fruits here. Please:

- [ ] Set the page title to something like "Internal error (HTTP code XXX)"
- [ ] Don't use an `<h2>` for the message, but a `<p>` element (it's not a header after all)

Those changes should remove the need for the extra CSS. In general, avoid creating CSS classes to a specific element (`section`), but simply use the class. Also, avoid class names that are too tied to the technical implementation (e.g. `hapi-error`), prefering the semantic meaning (e.g. `error-message`).

nightsh:
OK, noted, will do it now. I didn't want any name conflict with existing classes, that's why I went for those names. But I'll amend the code in a few minutes.
Regarding the page title, good catch.

Stay tuned, update coming soon :)

vitorbaptista:
Also remember to update the shinkwrap file

nightsh:
Removed the hapi-error dependency and did it myself via a request hook, we might need to think of a place to store custom functions instead of scattering them throughout the code. I can think of something, but if you have any suggestions, I'd be happy to hear them :)

vitorbaptista:
@nightsh You removed the `hapi-error` to be able to overwrite the `title` view attribute, right?

nightsh:
@vitorbaptista that's one reason, yes, but we also gained some extra flexibility by being able to adjust it more for our case(s).

vitorbaptista:
I should've pointed this to you before, but I'm using npm's website as an inspiration for many things in OpenTrials. I took a look at how they've done it. Check this out: https://github.com/npm/newww/blob/master/lib/error-handler.js

And then they just require that as part of the plugins (https://github.com/npm/newww/blob/ec7ed58a3406c2ebd7a121218b99349432d8d575/adapters/plugins.js).

Looks like a good pattern to follow. Don't bother with the different implementation, just where to put the code and how to load it.

nightsh:
Exactly what I had in mind. Is it OK with you if I create that structure? I would also extract `_addFlashMessagesToContext` into it if I ge there

vitorbaptista:
@nightsh Sure! :beers:

nightsh:
I updated the implementation (quite late, but I took some time to read more of the `newww` repo's code, as I'm sure it's going to pay off soon).

Please do not merge anything, I have to squash the latest commits together as they are quite redundant (kept them for reviewing purposes now, but at least one has to go).

lib/plugins.js
```diff
@@ -0,0 +1,8 @@
+'use strict';
+
+const plugins = {
+  addFlashMessagesToContext: require('./plugins/add-flash-messages-to-context'),
+  error: require('./plugins/error'),
```
vitorbaptista: `error` is a bit too general. Maybe change that to something more specific like `httpErrorHandler` or similar?

nightsh: Good catch!

nightsh:
Squashed and pushed!

lib/plugins/http-error-handler.js
```diff
@@ -0,0 +1,30 @@
+'use strict';
+
+function httpErrorHandler(request, reply) {
+  const response = request.response;
+  if (response.isBoom) {
+    let msg;
+    let statusCode = response.output.payload.statusCode;
+    let errorTitle = response.output.payload.error;
+    let errorMessage = response.output.payload.message;
```
vitorbaptista: Use `const` for variables that don't change. Also, you're not using the `msg` variable.

lib/plugins/http-error-handler.js
```diff
@@ -0,0 +1,30 @@
...
+
+    switch(statusCode) {
+      case 404:
+        errorMessage = 'Sorry, that page is not available.';
```
vitorbaptista: Could you explain your reasoning for overwritting the 404 error messages here? Why don't you use the default message?

lib/plugins/http-error-handler.js
nightsh: Ah, that's an artifact I'll remove now, weird that eslint didn't mention it! Also agreed with `const` usage

vitorbaptista: Agreed! I'd expect eslint to complain about it. Maybe it's somehow related to the new `lib/plugins` folder?

lib/plugins/http-error-handler.js
nightsh: I liked the idea implemented in hapi-error of producing friendlier messages for some errors, rather than having a dull "Not Found" in there. Which is only applicable when using them in templates, as they are perfectly fine for standard API usage. Does it make sense?

lib/plugins/http-error-handler.js
nightsh: Looking into it now.

nightsh: Turns out that my squash produced some undesired artifacts in the file, such as not using the shorthand assignments and various things previously fixed. Taking care of it now, good spot!

vitorbaptista:
Great job! :+1:

Issue #254, opened 2016-07-07
Display trials documents


vitorbaptista:
In our [mockups](http://opentrials.github.io/mockups/trial.html), we have a section for data:

![Data section in Trial details page](https://cloud.githubusercontent.com/assets/76945/16660607/059f9244-4467-11e6-8105-9d1d16ac5cb4.png)

Most of the data are trials' documents, but some are links (e.g. `ClinicalTrials.gov results`). We need to expose these documents through the API and add their UI.

# Requirements

* A trial can have multiple documents of the same type
* The documents are always shown in the same order in the Explorer

# Non Goals

* We're not adding the feature to flag wrong documents here

# Acceptance Criteria

- [x] Every trial displays a minimum set of documents in the Data section, regardless if we have them or not
- [x] Clicking on an available document will redirect you to its URL (can be a file or a link)
- [x] Clicking on an unavailable document redirects you to the contribute data page with the `trial_id` set to the current trial

# Questions

## Where to keep a list of the basic document types?

We need those types in the API, so we know which document is which. However, we also need them in the Explorer. Another complication is that we also want to group related documents together (e.g. regulatory documents, "paperwork", etc.). I can see three options:

1. Creating a new endpoint (`/v1/documents/types`) with a type list
2. Adding all documents to the trials response (`/v1/trials/{id}`), even the ones we don't have
3. Hardcoding

I'd rather avoid hardcoding where possible, so we can ignore option 3 for now. Options 1 and 2 are quite similar, but while option 1 adds this metadata in a new endpoint, option 2 embeds it inside the Trial response. I slightly prefer option 2, because it's easier and we avoid maintaining a new endpoint. There's a price to pay in increasing the API response size. However, considering that there aren't that many documents, and we're compressing the responses, it shouldn't matter much.

The response format I'm thinking is:

```javascript
{
    // ...
    documents: {
        regulatory: [
            {
                type: 'csr',
                url: 'http://somewhere.com/csr.pdf'
            },
            {
                type: 'epar_segment',
                // We don't have this document, so there' no URL
            },
        ],
        paperwork: [
            {
                type: 'blank_consent_form',
                url: 'http://somewhere.com/consent-form.pdf'
            },
        ],
        // ...
    }
}
```

In the Explorer-side, we'll have to have a dict to convert the documents types to strings. Should be trivial.

## How to handle these changes in the API database?

[...]

PR #262, opened 2016-07-13
[#254] Display Trials' documents in its details page

vitorbaptista:
To make it easier to display the documents data, I added a "presenter", which
can be thought of a View Model. It's useful to add and change the Trials'
attributes for display. Also, being a regular javascript file with no
dependencies to HapiJS or the DB, it's very easy to test.

In the future, a few other places could be refactored into presenters (e.g. the
trials' `contributeDataUrl`).

Depends on https://github.com/opentrials/api/pull/66


nightsh:
One small comment on the style, I would keep the exports / imports names consistent, as it's probably going to matter in the long run. (see `module.exports = decorate` vs. `const decorateTrial = require('../presenters/trial')`)

vitorbaptista:
Good point, @nightsh. Changed it so the codebase is a bit more "greppable" :smile:

PR #268, opened 2016-07-20
[#216] Add breadcrumbs to pages

### LINKED ISSUE #216: Add breadcrumbs to the pages
We already have the designs for it in our mockups:

http://opentrials.github.io/mockups/search-results.html
search results page breadcrumbs

http://opentrials.github.io/mockups/trial.html
trial details page breadcrumbs

# What to add in the breadcrumb?

## Search page

Just add the query string. That leaves the filters out for now. In the future, it would be nice to have a breadcrumb like:

> Search > `Depression` trials in `Brazil` or `UK`, `with published results` and `without discrepancies`.

But we can leave that for another task.

## Trial detail page

Add the name of the trial, but trim it after `x` characters so the breadcrumb doesn't appear too big

nightsh:
It fixes #216

### DIFF
--- a/views/partials/entity-details.html
+++ b/views/partials/entity-details.html
@@ -14,10 +14,12 @@
 
 {% extends "layouts/two-columns.html" %}
 
-{% block banner %}
-  {% block title %}<h1><small>{{ type | capitalize }}:</small>{{ title }}</h1>{% endblock %}
-  {% block subtitle %}<p>{{ subtitle }}</p>{% endblock %}
-  {% block bannerFooter %}{% endblock %}
+{% block title %}<h1><small>{{ type | capitalize }}:</small>{{ title }}</h1>{% endblock %}
+{% block subtitle %}<p>{{ subtitle }}</p>{% endblock %}
+{% block breadcrumbs %}
+  <li><a href="/">Home</a></li>
+  <li><a href="/search">Search</a></li>
+  <li><a>{{ title | truncate(35) }}</a></li>
 {% endblock %}
 
 {% block main %}
--- a/views/partials/search-form-fields.html
+++ b/views/partials/search-form-fields.html
@@ -24,12 +24,15 @@
 
   {% if trials %}
     <ul class="breadcrumb">
+      <li><a href="/">Home</a></li>
+      <li><a>Search</a></li>
+      {% if query.q %}
       <li>
-        <a href="/search">Search</a>
-      </li>
-      <li>
-        Found {{ trials.total_count | formatNumber }} trial{% if trials.total_count != 1 %}s{% endif %}
-      </li>
+        <a>
+          {{ query.q }} ({{ trials.total_count | formatNumber }}
+          trial{% if trials.total_count != 1 %}s{% endif %})
+        </a></li>
+      {% endif %}
     </ul>
   {% endif %}
 </div>
--- a/views/trials-details.html
+++ b/views/trials-details.html
@@ -6,7 +6,14 @@
 {% endif %}
 {% endmacro %}
 
+{% block breadcrumbs %}
+  <li><a href="/">Home</a></li>
+  <li><a href="/search">Search</a></li>
+  <li><a>{{ title | truncate(35) }}</a></li>
+{% endblock %}
+
 {% block bannerFooter %}
+
   <ul class="overview-list">
     {% if trial.registration_date %}
     <li>
--- a/assets/styles/ot-scss/_banner.scss
+++ b/assets/styles/ot-scss/_banner.scss
@@ -64,6 +64,12 @@
         }
       }
     }
+
+    li:last-child {
+      a:after {
+        content: '';
+      }
+    }
   }
 
   .overview-list {
--- a/handlers/about.js
+++ b/handlers/about.js
@@ -6,4 +6,6 @@ function about(request, reply) {
   });
 }
 
-module.exports = about;
+module.exports = {
+  handler: about,
+};
--- a/handlers/admin/data-contributions.js
+++ b/handlers/admin/data-contributions.js
@@ -8,11 +8,11 @@ function editDataContribution(request, reply) {
   const data = {
     approved: request.payload.approved,
     curation_comments: request.payload.curation_comments,
-  }
+  };
 
   return new DataContribution({ id: request.params.id })
     .save(data, { patch: true, require: true })
-    .then((dataContribution) => {
+    .then(() => {
       request.yar.flash('success', 'Data Contribution updated successfully.');
       return reply.redirect('/admin/data-contributions');
     })
--- a/handlers/condition.js
+++ b/handlers/condition.js
@@ -24,4 +24,6 @@ function conditionsDetails(request, reply) {
   });
 }
 
-module.exports = conditionsDetails;
+module.exports = {
+  handler: conditionsDetails,
+};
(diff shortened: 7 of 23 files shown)


nightsh:
@vitorbaptista I'm available anytime to discuss some changes I made here. Also, I didn't run eslint yet, I'll wrap it up now (a bit of a premature/preview push).

nightsh:
I still can't tell where is the decreased coverage, can you spot it anywhere?

vitorbaptista:
Big :+1: for adding `hapi-to`. I'm not sure about the actual breadcrumbs implementation, though. I think it's a bit overkill for the few breadcrumbs we want to add...

vitorbaptista:
About the coverage, check the "All" tab in https://coveralls.io/builds/7084109

nightsh:
1. The coverage: I've seen it now, fix will be on its way soon.

2. The breadcrumbs: we still had to get some custom behaviour in there, it probably would've been dirtier and not very different, just bundled within a handler. Or more.

3. `hapi-to`: please note I changed the routes' exports to be all the same (some were exporting only the handler). Which brings me to the next question: should we rename `handlers` directory to `routes`, now that the files inside are all exporting routes configs and not only handlers?

nightsh:
I'm fighting the coverage thingy here, but I really don't see why I should add a test for the thing it highlights there.

vitorbaptista:
@nightsh Does your code support more than 2 breadcrumb levels (like `Search > Trial X > Record Y`)? I would also like to see breadcrumbs like `Home > Contribute data`, `Home > About` for the basic pages.

I'm still not very fond of the automatic breadcrumbs creation. I think it'll end up gathering a lot of corner cases and different flows moving forward. I understand this will probably mean that we won't be able to have breadcrumbs like `Search > Depression > Trial X`, but I can live with that.

Also, remember that the last breadcrumb (i.e. the current page) shouldn't have a link (you can disable it with CSS using `pointer-events`). And please keep the handlers' `id` as the first key in the dict definitions, as in:
```javascript
// no
{
  handler: foo,
  id: 'foo'
}

// yes
{
  id: 'foo',
  handler: foo,
}
```

nightsh:
Regarding multiple levels, it should, where can I find such a nesting?

If it fails and starts gathering corner cases, which is possible, I will probably clean up the thing by placing them in a more "manual" approach. Won't be a hard thing to do at any given moment.

On all the other pages that are not Search, I missed the ":Home" link, right. fixing it now.
Same for the `id` issue, I'll make it first, thanks.

2 more things:

* I intentionally left the link on the current breadcrumb, so we'd provide a quick way to get a permalink of it. There's one extra line of code that does that, so it'd be a breeze to remove the `href` entirely ;)
* if you look at `routes/index.js`, you'll see that the files in `handlers` are now route configs with handlers embedded in them, thus questioning a bit the name of the `handlers` directory itself.

nightsh:
As discussed on Gitter, moving to hardcoded breadcrumbs to avoid a cluster of conditions that would treat all the corner cases (wow, I had no idea the special cases count would rise up so high for such a simple thing).

Will keep: `hapi-to`, the routes configuration, the styling on the breadcrumbs.
Will remove: the breadcrumbs middleware + its libraries and tests + template iterations.
Will add: static breadcrumbs in the templates related to the top banner.

nightsh:
quick note, don't merge before I squash everything into one piece

views/trials-details.html
```diff
@@ -7,6 +7,15 @@
...
 {% block bannerFooter %}
+
+  <ul class="breadcrumb">
+    {% block breadcrumbs %}
+      <li><a href="/">Home</a></li>
+      <li><a>Trials</a></li>
+      <li><a>{{ title | truncate(35) }}</a></li>
+    {% endblock %}
+  </ul>
+
```
vitorbaptista: I think you can avoid redefining the breadcrumb `ul` here by using `{{ super() }}` or, even better, getting the `breadcrumbs` block outside of the `bannerFooter` on the layout.

vitorbaptista:
I know this is still WIP, but just to be sure we don't forget it. In general, every breadcrumb should be a link (except the last one, which is the current page). As we don't have pages to list (for example) organisations, we can't do a breadcrumb like `Home > Organisations > J&J`. In that case, having a `Home > Search > J&J` would be more useful IMHO.

Overall, breadcrumbs seem not to fit well with our pages structure, so they end up a bit weird. We'll see how things go whenever this is merged.

nightsh:
This is ready to merge as far as I'm concerned.

* breadcrumbs are statically placed into templates
* above discussion and line comments are taken into consideration
* some small code improvements
* everything squashed into one piece commit

views/partials/entity-details.html
```diff
@@ -14,10 +14,12 @@
...
-{% block banner %}
-  {% block title %}<h1><small>{{ type | capitalize }}:</small>{{ title }}</h1>{% endblock %}
-  {% block subtitle %}<p>{{ subtitle }}</p>{% endblock %}
-  {% block bannerFooter %}{% endblock %}
+{% block title %}<h1><small>{{ type | capitalize }}:</small>{{ title }}</h1>{% endblock %}
+{% block subtitle %}<p>{{ subtitle }}</p>{% endblock %}
+{% block breadcrumbs %}
+  <li><a href="/">Home</a></li>
+  <li><a href="/search">Search</a></li>
+  <li><a>{{ title | truncate(35) }}</a></li>
```
vitorbaptista: I'm not sure as I haven't run the code, but you might be missing a `{% block banner %}` and a `{% block bannerFooter %}{% endblock %}` here?

views/partials/search-form-fields.html
```diff
@@ -24,12 +24,11 @@
...
-      <li>
-        <a href="/search">Search</a>
-      </li>
-      <li>
-        Found {{ trials.total_count | formatNumber }} trial{% if trials.total_count != 1 %}s{% endif %}
-      </li>
+      <li><a href="/">Home</a></li>
+      <li><a>Search</a></li>
+      {% if query.q %}
+        <li><a>{{ query.q }}</a></li>
```
vitorbaptista: Add the number of trials as well. Maybe something like `{{ query.q }}({{ trials.total_count | formatNumber }} trial(s))`

Also, make sure the breadcrumb for the current page isn't clickable

views/partials/entity-details.html
nightsh: Nope, I'm only redefining a block that is already positioned and defined in the layout this view extends. In this case, the `breadcrumbs` block.

views/partials/search-form-fields.html
nightsh: I've done it a bit differently to avoid the need for "(s)" which looked weird in this context, creating a nested parenthesis.

vitorbaptista:
I added a few modifications and merged with https://github.com/opentrials/opentrials/commit/ae8e010b30b3b98426a8d053829cb1964e0701a4. My changes were:

* Rename `breadcrumbs` block to `breadcrumb` to be consistent with the class name
* Removed `<a>` elements from the current page in the breadcrumb (they shouldn't be clickable, so adding this element is irrelevant). This change also simplifies the CSS.
* Fixed an unrelated bug with `formatNumber` (it wasn't working with the number `0`)
* Added the number of trial to the search results breadcrumb, even when there's no query, so the user can see how many trials we have in total

PR #322, opened 2016-08-23
[#311] Form's "Clear" button clears all inputs

### LINKED ISSUE #311: Search's "Clear" button should clear everything, including the actual query string
For example, go to http://explorer.opentrials.net/search?q=breast+cancer&location=United+Kingdom and click on the `Clear` button beside the `Search` button. Notice that it'll remove the `location` filter, but will still keep the `breast cancer` query.

To solve this, you must change the javascript at https://github.com/opentrials/opentrials/blob/master/assets/js/index.js#L62-L70. Instead of clearing just the fieldset, change it to clear all fields from the containing `form`. Make sure you restrict yourself to the current form, and not any other that might be on the page.

vitorbaptista:
I've also changed the button from "#clear-fieldset" to ".clear-form"

Fixes #311.


views/partials/search-form-fields.html
```diff
@@ -183,7 +183,7 @@
   <div class="controls">
     <button type="submit"><span>Search</span></button>
     {% if advancedSearchIsVisible %}
```
nightsh: This line shouldn't be necessary, since the whole container is invisible if `advancedSearchIsVisible` is false. Also, it doesn't show up on my machine the way it's built right now.

nightsh:
On my machine the "Clear" button triggers a browser refresh, I wonder if we could avoid that and just clear the fields. I assume that is the intended behaviour anyway.

views/partials/search-form-fields.html
vitorbaptista: The container is hidden by default, but the user can click on "Advanced Search" and see it. Maybe it's overkill, and it's OK to show the "Clear" button even if there's nothing to clear, but it doesn't hurt considering it's already here.

vitorbaptista:
@nightsh The "Clear" button is supposed to do a refresh to get the updated search results. This way the user doesn't need to click on "Clear" and then "Search".

views/partials/search-form-fields.html
nightsh: I can't see it unless I remove the condition in the template, weird... Is it showing properly on your setup?

vitorbaptista: It'll only appear if you already have some advanced search filters in place (e.g. http://explorer.opentrials.net/search?location=Romania)

nightsh: OK, it makes sense now. I was somehow expecting it to show up after setting the filters, but before hitting "Search". Sorry for the noise, I overlooked a few bits of code here :-)

nightsh:
IMO it's good to go into master.

Issue #287, opened 2016-07-27
Write API documentation


vitorbaptista:
We have some very crude docs on http://api.opentrials.net/v1/docs/, but we need to improve them a lot.

PR #349, opened 2016-08-29
Minor UI tweaks

vitorbaptista:
I wanted to get us closer to the mockup's UI:

mockup UI

The biggest change here is to move the "Data" documents to the upper part of the page, and moving them as top-down instead of sideways. I prefered not to do so now because I think it'll end up looking a bit odd when displaying trials with "unbalanced" documents. Having it sideways is more resilient in my opinion.

I also changed the sources list to link directly to the source URL, and not our record's page. This means that there's no way to access the record's pages now. We can leave it as is for now but, if this becomes our intended solution, we should remove the handlers/views that aren't needed anymore. Also, I didn't rename "Sources" to "Registries" because we have sources that aren't registries.

### DIFF
--- a/test/e2e/search.js
+++ b/test/e2e/search.js
@@ -26,9 +26,8 @@ describe('(e2e) search', function() {
     driver.get(SERVER_URL);
     driver.findElement(By.css('.search-bar input')).submit();
     driver.findElement(By.css('.search-results .title a')).click();
-    driver.findElement(By.css('.actions .download')).click();
-    return driver.getPageSource()
-      .then((body) => should(body).not.containEql('"error"'))
+    return driver.getCurrentUrl()
+      .then((url) => should(url).match(/\/trials\//));
   });
 
   it('should work with all search filters enabled', () => {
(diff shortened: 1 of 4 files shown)


roll:
May be we could be able to remove `database.records.source_data`? It's more than 50% of our database size.. Messing with sql dumps downloading etc.

nightsh:
👍 
Tested locally, all works as described. It's ready to go AFAICT.

I also move the discussion to #350, so we can still talk about it after this PR is closed.

davidmiller:
Hi team!

Now that there's no link to a trial's API endpoint, I had to spend 10 minutes reading the source code to get to the swagger documentation :(

A link somewhere on the site would have been neat.

vitorbaptista:
Hey @davidmiller, sorry for the hassle. We're just starting to work on our API documentation at https://github.com/opentrials/opentrials/issues/287 in preparation for our [hack day](https://www.eventbrite.com/e/opentrials-hack-day-tickets-27046834811) in the beginning of October.

If you're interested, I would suggest you to subscribe to the issue #287 so you're notified when our docs are online. We're also on Gitter at https://gitter.im/opentrials/chat if you have any questions or issues. I'd be curious on any usages of our API and data :smile:

Issue #354, opened 2016-08-30
Update the fields we use to calculate discrepancies


vitorbaptista:
As per https://github.com/opentrials/opentrials/issues/234#issuecomment-235896623, we should consider only:

* target_sample_size
* gender
* status
* recruitment_status
* has_published_results

We're not adding `study_phase` as well because their values aren't normalized yet (see https://github.com/opentrials/opentrials/issues/353), so we would end up with many false positives.

vitorbaptista:
Fixed by https://github.com/opentrials/opentrials/pull/355

PR #355, opened 2016-08-30
[#354] Update the fields we use to calculate discrepancies

vitorbaptista:
@nightsh I'm merging this before your review as it's a trivial change and I'd like to calculate the new discrepancies today. Please do review when you see this and we can change as needed.


nightsh:
👍

Issue #281, opened 2016-07-27
Expose license and attribution information via API and UI


vitorbaptista:
It should be possible to know the license and where the contents of each page come from. We already have the attribution data on the Trials page in the form of the "Sources" list, and in the API, where every Trial's Records have a `source` field pointing to where they come from. We're then missing license information.

@pwalsh Considering that our sources might use custom license (e.g. [EU CTR](https://www.clinicaltrialsregister.eu/disclaimer.html)), I wonder how we would show them to the user. Maybe it's enough to just point out where the data comes from (i.e. attribution)? We can then have a new page in OpenTrials where we list all our sources and link to their licenses (if they have one).

# Tasks

- [x] Add `url` and `terms_and_conditions_url` attribute to `sources` table
- [x] Find where the data's license is defined in each source, and change the corresponding collector/processor to add it to the `sources.terms_and_conditions_url` column. Same for `sources.url`.
- [ ] Create new "Terms and Conditions" page with the list of sources, their URLs and their license's URLs

# Data

```csv
id,url,terms_and_conditions_url
actrn,http://www.anzctr.org.au,http://www.anzctr.org.au/Support/Terms.aspx
euctr,https://www.clinicaltrialsregister.eu,https://www.clinicaltrialsregister.eu/disclaimer.html
fda,http://www.fda.gov,
fdadl,https://open.fda.gov,https://open.fda.gov/terms/
gsk,http://www.gsk.com,http://www.gsk.com/en-gb/terms-of-use/
hra,http://www.hra.nhs.uk,http://www.hra.nhs.uk/terms-conditions/
icdcm,https://www.cms.gov/Medicare/Coding/ICD10/index.html,
icdpcs,https://www.cms.gov/Medicare/Coding/ICD10/index.html,
ictrp,http://www.who.int/trialsearch/,http://www.who.int/ictrp/search/download/en/
isrctn,http://www.isrctn.com/,http://www.isrctn.com/page/terms
jprn,http://rctportal.niph.go.jp/en/,http://rctportal.niph.go.jp/en/agreement
nct,https://clinicaltrials.gov,https://clinicaltrials.gov/ct2/about-site/terms-conditions
pfizer,http://www.pfizer.com/research/clinical_trials,http://www.pfizer.com/general/terms
pubmed,http://www.ncbi.nlm.nih.gov/pubmed,https://www.ncbi.nlm.nih.gov/home/about/policies.shtml
[...]

PR #386, opened 2016-09-07
[#281] Add terms of use page

vitorbaptista:
@nightsh @benmeg I assigned both of you here to review the tech parts and the actual text.

@benmeg You can see the page at https://github.com/opentrials/opentrials/blob/add-terms-and-conditions/views/terms-of-use.html

Depends on https://github.com/opentrials/api/pull/81

Fixes opentrials/opentrials#281

### DIFF
--- a/agents/sources.js
+++ b/agents/sources.js
@@ -1,0 +1,13 @@
+'use strict';
+
+const opentrialsApi = require('../config').opentrialsApi;
+
+function listSources() {
+  return opentrialsApi
+    .then((client) => client.sources.list())
+    .then((response) => response.obj);
+}
+
+module.exports = {
+  list: listSources,
+};
--- a/test/handlers/terms-of-use.js
+++ b/test/handlers/terms-of-use.js
@@ -1,0 +1,47 @@
+'use strict';
+
+const server = require('../../server');
+
+describe('terms-of-use handler', () => {
+  describe('GET /terms-of-use', () => {
+    describe('API is OK', () => {
+      const sources = JSON.parse(JSON.stringify([
+        fixtures.getSource(),
+        fixtures.getSource(),
+      ]));
+      let response;
+
+      beforeEach(() => {
+        apiServer.get('/sources').reply(200, sources);
+      });
+
+      it('is successful', () => (
+        server.inject('/terms-of-use')
+          .then((response) => response.statusCode.should.equal(200))
+      ));
+
+      it('uses the "terms-of-use" template', () => (
+        server.inject('/terms-of-use')
+          .then((response) => response.request.response.source.template.should.equal('terms-of-use'))
+      ));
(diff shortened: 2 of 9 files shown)


agents/sources.js
```diff
@@ -0,0 +1,13 @@
...
+const opentrialsApi = require('../config').opentrialsApi;
+
+function listSources() {
+  return opentrialsApi
+    .then((client) => client.sources.list())
+    .then((response) => response.obj);
+}
+
+module.exports = {
+  list: listSources,
```
nightsh: Same as for the API PR, I would go with `getSources` as `operationId`, then export it as it is, using the new ES6 syntax ;)

test/fixtures/api/swagger.yaml
```diff
@@ -294,6 +294,23 @@ paths:
           schema:
             $ref: "#/definitions/ErrorResponse"
 
+  /sources:
```
nightsh: This will continue being a source of pain, can't we `gitignore` this file and grab it from the git repo if it's missing? It's annoying to always have to replicate API's `swagger.yml` and also easy to forget

test/handlers/terms-of-use.js
```diff
@@ -0,0 +1,47 @@
...
+          .then((response) => {
+            const context = response.request.response.source.context;
+            context.sources.should.deepEqual(sources);
+          })
+      });
+    });
+
+    describe('API is not OK', () => {
+      it('returns error 502', () => {
+        apiServer.get('/sources').reply(500);
```
nightsh: 👍

test/fixtures/api/swagger.yaml
vitorbaptista: Yeah, it's quite annoying. However, I would rather not require a HTTP request to run the tests. It's also somewhat important to keep it together so if we use a tool like `git-bisect` to find for a bug, the tests will always pass, even if the current `swagger.yaml` is totally different from what it was at the time the test was written.

agents/sources.js
nightsh: @vitorbaptista what do we do with this one?

vitorbaptista: Depends on what get merged first, if this or https://github.com/opentrials/api/pull/80. As this is an easier PR, I would prefer to merge it before and then do the fixes as part of other PR, but we'll see.

nightsh:
@vitorbaptista can we merge this one? It's OK from my point of view, and it's currently blocking my Swagger PRs since it contains breaking changes (I'd rather make the breaking changes myself afterwards)

nightsh:
@vitorbaptista want me to take this one and prepare it for merge? I can help.

vitorbaptista:
@nightsh Sure, please do it :)

PR #391, opened 2016-09-14
Don't swallow errors when starting server

vitorbaptista:
For more details, check
http://jamesknelson.com/are-es6-promises-swallowing-your-errors/

Without this patch, if there're any exceptions thrown while starting the server, the process will just hang. No messages will be shown at all, so it's quite hard to debug.


nightsh:
Ah, good `catch`! 👍

Issue #305, opened 2016-07-31
Improve the UI of the search results page?


roll:
Here is our search results page: http://explorer.opentrials.net/search

Comparing to other sites:
- https://mytomorrows.com/en/conditions/580/adenocarcinoma/trials/page/1/?country=RU&gender&intervention&sortBy=alphabetical_order
- http://www.takedaclinicaltrials.com/browse/?sub_conditions=&compound=&country=&age=&search_terms=cancer&protocol_id=&nct_id=&study_type=
- http://www.pfizer.com/research/clinical_trials/find_a_trial?group[cancer]=1&condition[cancer]=cancer&recr=1
- http://apps.who.int/trialsearch/Default.aspx
- etc

I should say ours is not looking as a leader. For example critical information like `recruitment_status` is missing. I think we need to invest some time into it before the launch (we even have things like `Internetions: <blank line>`)

nightsh:
My 2 cents on the issue:

We're not far from having a much better search page. The current structure allows for a nicer presentation without having to modify much of it. I personally liked takedaclinicaltrials the most.

Some ideas to start with:

* find a way to add more data to the results previews without bloating the UI (icons with tooltips?)
* a friendlier (maybe with a smaller footprint)  "refine search results" thingy would probably be nice
* visually (and discretely) isolate the search results, either by a small separator, (alternate?) backgrounds or something else
* whatever we do, I'd avoid using colors and/or color codes. Just look [at this page](http://www.takedaclinicaltrials.com/browse/?sub_conditions=&compound=&country=&age=&search_terms=rectal+cancer&protocol_id=&nct_id=&study_type=) and see how crowded it looks

smth:
As part of this I'd  like to take a look at how we present the "advanced" fields, as mentioned here https://github.com/opentrials/opentrials/issues/347

smth:
Here's a mockup with the "advanced" filters moved into this page, amongst some smaller tweaks. Would this resolve the concerns here?

![opentrials_08_search_01](https://cloud.githubusercontent.com/assets/1375708/18275597/9f34f9dc-743f-11e6-9a4c-75b54b8a54b1.jpg)

vitorbaptista:
@smth Looks great! A few comments:

[...]

PR #392, opened 2016-09-19
Search results improvements

smth:
Addresses most of #305


assets/js/vendor/index.js
```diff
@@ -3,3 +3,12 @@
 require('nodep-date-input-polyfill');
 
 require('jquery.mmenu');
+
+require('tooltipster');
+
+$(document).ready(function() {
+  $('.hastip').tooltipster({
+     trigger: 'click',
```
vitorbaptista: Can we enable hover as well?

vitorbaptista: Also, move it to https://github.com/opentrials/opentrials/blob/master/assets/js/index.js instead. This file should only contain the `requires`.

vitorbaptista:
I also found a pretty minor bug on the advanced search. To reproduce, do:

1. Open the search page with a big resolution (to get the "desktop version")
2. Reduce the screen size until you get the mobile version.
3. Click on `Advanced search` to expand the options, then click again to remove them
4. Increase the screen size to get back to the desktop version

The advanced search filters will be hidden, and the `Advanced search` button won't be clickable (so you can't expand them).

If that's difficult to fix, I'm happy to just create an issue.

advanced search bug

vitorbaptista:
Also, please check the Travis build. There's a minor error with linting the JS code.

assets/js/vendor/index.js
smth: We can have hover for mouse and tap for touch, if that's what you're alluding to. Having hover and click for mouse seems unnecessary.

smth:
@vitorbaptista What do you mean by odd in regards to the filters?

I was aware of that issue with resizing, but it seems like a pretty unlikely scenario to me.

Issue #327, opened 2016-08-24
Remove differences between primary and secondary identifiers


vitorbaptista:
As per #323, we don't need to treat them differently. We must then update the explorer, api, and maybe collectors/processors to treat all identifiers the same. The UI on the Explorer should look like:

![trial identifiers list](https://cloud.githubusercontent.com/assets/76945/17903923/292c21f0-6965-11e6-9fd7-2ca7c1f66839.png)

If possible, try to avoid needing multiple API calls to populate a single page. However, as the Explorer is "only" a user of the API, make sure we're not changing it too much to work for our specific use case (it's hard to keep the balance). As an inspiration, check how the `records` are embeded in the `trial` API response. An example of what this API response could look like:

```json
{
	"records": [
		{
			"source_id": "nct",
			"source_url": "http://clinicaltrials.gov/NCTXXXXX"
		},
		{
			"source_id": "euctr",
			"source_url": "http://example.org/EUCTRXXXX"			
		}
	],
	"sources": {
		"nct": {
			"id": "nct",
			"name": "ClinicalTrials.gov",
		},
		"euctr": {
			"id": "euctr",
			"name": "EU Clinical Trials Registry",
		},
	},
	"identifiers": {
		"nct": "NCTXXXXXXX",
		"euctr": "EUCTRXXXXX",
	}
}
```

benmeg:
Related to this - we're not displaying all 'secondary' IDs: #329

roll:
@benmeg 
also related - https://github.com/opentrials/opentrials/issues/360

PR #397, opened 2016-09-20
[#327] Remove differences between primary and secondary identifiers

vitorbaptista:
Fixes opentrials/opentrials#327


views/trials-details.html
```diff
@@ -191,29 +191,18 @@ <h2 class="disabled">{{ name }}</h2>
...
       </dt>
       <dd>
         {{ trial.id }}
       </dd>
 
-      {% if primaryId %}
+      {% for identifier in trial.identifiers %}
       <dt>
-        Primary ID:
+        {{ identifier.name }}
```
nightsh: Shouldn't we uppercase this? AFAIK there are only (mostly?) abbreviations in.

vitorbaptista: The names should be things like "ClinicalTrials.gov", "EU CTR", etc. The name should be corrected in the DB-level.

nightsh:
👍

Issue #393, opened 2016-09-19
Setup permanent links for takedown request and flag error pages.


vitorbaptista:
They're currently direct links to Google Docs. However, those links aren't stable, so we should create a `explorer.opentrials.net/takedown` and other for the error pages. For now, they'll just redirect to Google Docs, but that allow us to change them later without breaking other links.

PR #408, opened 2016-09-22
[#393] Encode URLs before sending to Google Forms

vitorbaptista:
Otherwise they'll mess up the other query parameters

opentrials/opentrials#393


nightsh:
Sure, let's see it in `master`! 👍

Issue #412, opened 2016-09-27
Sort source names in trial details page


vitorbaptista:
You can sort by `id` as well, as it will probably be easier. This is to organize a bit when there're multiple trials with the same name, for example:

![unordered trial sources](https://cloud.githubusercontent.com/assets/76945/18885935/a69d8566-84e5-11e6-9b6d-d21c02293826.png)

PR #413, opened 2016-09-27
[#412] Sort sources by name on trial details page

nightsh:


nightsh:
spectacle lh6774

presenters/trial.js
```diff
@@ -71,6 +71,7 @@ function decorateIdentifiers(identifiers, sources) {
 }
 
 function decorateTrial(trial) {
+  trial.records = _.sortBy(trial.records, 'source_id');  // eslint-disable-line no-param-reassign
```
vitorbaptista: Instead of changing the records here, do it a few lines later,  like:

```javascript
    {
      documents: decorateDocuments(trial.documents || []),
      identifiers: decorateIdentifiers(trial.identifiers || {}, trial.sources || {}),
      records: // ...
    }
```

Check also if `trial.records` never is undefined. To be safe, it might be better to use `trial.records || []`.

Also, please add a test for this.

nightsh: Ah, you're right, this looks much better. Will update.

test/presenters/trial.js
```diff
@@ -56,6 +56,69 @@ describe('trial presenter', () => {
...
+          "url": "http://localhost:10010/v1/trials/ee6883be-3414-456c-82aa-91d802a83c98/records/91d1e3eb-e512-47ca-88b9-bb1013e55ef0",
+          "source_url": "https://clinicaltrials.gov/ct2/show/NCT01193257",
+          "updated_at": "2016-02-07T03:26:01.242Z"
+        },
+      ]);
+    });
+
+    it('works if there are no records', () => {
+      const trialAttributes = {
+        records: []
```
vitorbaptista: This should be `records: undefined` (or simply `trialAttributes = {}`)

presenters/trial.js
```diff
@@ -56,6 +56,10 @@ function decorateDocuments(documents) {
   return documentsByCategory;
 }
 
+function decorateRecords(records) {
+  return _.sortBy(records || [], 'source_id');
```
vitorbaptista: No need to do `records || []` here, as you're already doing it when calling the function in line 84.

test/presenters/trial.js
```diff
@@ -56,6 +56,69 @@ describe('trial presenter', () => {
...
+            "source_id": "nct",
+            "url": "http://localhost:10010/v1/trials/ee6883be-3414-456c-82aa-91d802a83c98/records/91d1e3eb-e512-47ca-88b9-bb1013e55ef0",
+            "source_url": "https://clinicaltrials.gov/ct2/show/NCT01193257",
+            "updated_at": "2016-02-07T03:26:01.242Z"
+          },
+        ]
+      };
+      const trial = trialPresenter(trialAttributes);
+
+      should(trial.records).deepEqual([
```
vitorbaptista: As a suggestion, you could do something like:
```javascript
const recordsSources = trial.records.map((record) => record.source_id);
should(recordsSources).deepEqual(['euctr', 'ictrp', 'nct']);
```

This is because you don't care (on this test) about the other fields, just that the order is correct. This makes the test a bit more resistant to changes in the `records` structure.

nightsh: I thought about it myself (i.e. ignoring the other fields) but I put a real sample in just to make sure in a very paranoid way it holds against real data. Not entirely justified, it just felt safer :grinning: 

You think I should change it to just ignore the rest of the JSON fluff in `records`?

vitorbaptista: When testing, a good practice is to assert a single thing per test. Some people say that a test should actually have a single `assert`, and others say that a test should test a single `concept` (that can have multiple asserts). I'm in the latter team. In this test, you're checking if the records are ordered by `source_id`. To do that, the only thing you need is to check the `source_id`s of the records.

This makes it easier to follow another testing best practice: there should be a single reason for a test to fail. The way it's written now, the test can fail if we, for example, parse the `updated_at` date as a `datetime` object. This has nothing to do with the sorting, which is what you're testing here.

nightsh: I integrated your changes, thanks for the explanations and suggestions ;)

Issue #426, opened 2016-09-30
Setup error logging


vitorbaptista:
Probably using Sentry and https://github.com/bendrucker/hapi-raven

PR #439, opened 2016-10-03
[#426] Setup Sentry and improve logging

vitorbaptista:


nightsh:
There seems to be a problem if you have Sentry installed, but no DSN in your `.env`. Can you check in your environment as well? On my machine it crashes the explorer (`[ops]`).

---

On a more positive note:
> Coverage increased (+0.3%) to 100.0%

Woohoo! 🎉

vitorbaptista:
Talked with @nightsh and not sure what issue he's seeing. As everything is working when setting `SENTRY_DSN` for our production instance, and the code shouldn't do anything at all unless you actually have a `SENTRY_DSN` env variable set, we won't spend more time debugging this.

nightsh:
It was my environment having a zombie listening to the development API port. Silly though. Sorry for the noise :)

Issue #446, opened 2016-10-04
Registry links on discrepancy pages don't link to source


benmeg:
e.g.

http://explorer.opentrials.net/trials/f9350f32-5462-4119-bf29-4e1a3fcef698/discrepancies

ISRCTN and WHO ICTRP (column headings) link to the same page, rather than the source of the trial data

PR #448, opened 2016-10-04
[#446] Fix bug where discrepancies' sources links were undefined

vitorbaptista:
The problem is that `discrepancies` doesn't have a `source_url`. We have to
take it from the `records` array.

Fixes opentrials/opentrials#446


nightsh:
👍 all good!

Issue #417, opened 2016-09-28
Lists are appearing with bullets where they previously haven't


vitorbaptista:
For instance, in the publications list at http://explorer.opentrials.net/trials/763d04aa-651a-4b0b-ae97-f1e4a2edc5f3:
![image](https://cloud.githubusercontent.com/assets/76945/18911648/2d2ebf42-8576-11e6-86c6-e34fbb111e72.png)

And the publication authors at http://explorer.opentrials.net/publications/d803a5ab-e0e0-4a4f-89ee-1bccc32ba81c:
![image](https://cloud.githubusercontent.com/assets/76945/18911658/405752aa-8576-11e6-9a37-018c2edaea1d.png)

This is probably a regression caused by https://github.com/opentrials/opentrials/issues/236

PR #475, opened 2016-10-06
[#417] Remove bullets from side unordered lists

georgiana-b:
opentrials/opentrials#417


vitorbaptista:
The bullets were removed, but the lists still have the padding, which makes them look broken:

lists without bullets but with margin

vitorbaptista:
Merged as 635396aaf0ef92ed6167049a9e4a54b2d5979f05. Also added some bottom marging.

Issue #437, opened 2016-10-03
Expose Cochrane risks of bias


nightsh:
Currently we have a processor writing risks of bias for trials in the API database. There's also a migration, the corresponding models and some tests, and everything is almost done.

What needs to be done in order to be able to use this data is:

- [x] Wrap up the work on PR opentrials/api#87 (and optionally merge it)
- [x] Ensure the API is serving the risks of bias as a JSON formatted property on trial (not on the summary, just the details)
- [x] Add the new specs to Swagger
- [x] Sync Swagger on explorer
- [x] Consume the updated API response and present it to the users.

The mockup for trial details, including the new section, is here: http://opentrials.github.io/mockups/trial.html
(the "Methodological rigour" section on details)

nightsh:
Ping @vitorbaptista, make sure you don't create a duplicate of this issue :)

vitorbaptista:
Fixed

PR #479, opened 2016-10-07
[#437] Improve the risk of bias messages

vitorbaptista:
Without this patch, the message looks like:

> The methodological rigour of this trial was scored in a systematic review, where it was scored as being "low risk" of bias for "sequence generation" "unclear" for "allocation concealment" "low risk" of bias for "blinding (performance and/or detection)" "low risk" of bias for "attrition" "low risk" of bias for "reporting" "high risk" of bias for "other biases" .

With this, it becomes:

> The methodological rigour of this trial was scored in a systematic review, where it was scored as being at "low risk" of bias for "sequence generation", "unclear" for "allocation concealment", "low risk" of bias for "blinding (performance and/or detection)", "low risk" of bias for "attrition", "low risk" of bias for "reporting", and "high risk" of bias for "other biases".

Note the lack of commas and the whitespace before the final period in the previous message.

opentrials/opentrials#437


nightsh:
Awesome, thanks! Let's merge it 👍

PR #496, opened 2016-10-17
Limit Node memory usage to 512 MB

vitorbaptista:
This is what we have in Heroku, and it avoids us receiving exceptions regarding
to our memory quota being exceeded (by default Node uses up to 1.5 GB).


vitorbaptista:
Merging this, as it's a small change and is already being used in the API (so we know it works).

PR #548, opened 2016-11-03
Set data_contributions.approved as nullable and add default values to the tables with timestamps

vitorbaptista:
Having `data_contributions.approved` as nullable allow us to see which contributions weren't curated yet (i.e. neither approved nor rejected).


georgiana-b:
:+1:

helpers/escape-elastic-search.js
```diff
@@ -2,7 +2,7 @@
 
 function escapeElasticSearch(query) {
   return query
-    .replace(/[\*\+\-"~\?^{}\(\):!\/[\]\\]/g, '\\$&')
+    .replace(/[*+\-"~?^{}():!/[\]\\]/g, '\\$&')
```
vitorbaptista: This example is a bit tricky. See:

```
$ node
> 'yes\ no'
'yes no'
> 'yes\\ no'
'yes\\ no'
> 'yes\ no' === 'yes no'
true
```

In other words, the `\` in `'yes\ no'` is escaping the following whitespace. As `'\ '` has no meaning in JavaScript, `'\ ' === ' '` is true.

Issue #500, opened 2016-10-17
Create API endpoint for searching the FDA data


vitorbaptista:
Maybe this could be a general endpoint for searching the `documents` table. Need to investigate more.

vitorbaptista:
Blocked on #499

PR #551, opened 2016-11-03
[#500] Remove FDA documents from trials.documents

vitorbaptista:
They should be treated as different entities, as we'll display them
differently.

opentrials/opentrials#500


vitorbaptista:
@nightsh Need to merge this to be able to push the code to production, and need to push the code to prod to start extracting the file's text (which will take a looong time). Please take a look later :+1:

nightsh:
I am actually using this now, quite a welcome change, so I guess thanks are in order :stuck_out_tongue_winking_eye:  

Code is OK, I'm glad you merged it :+1:

Issue #560, opened 2016-11-10
Display FDA documents in explorer


nightsh:
- [x] implement mockup
- [x] decorate the trial presenter with the FDA docs
- [x] show the data in template
- [x] pull a small content snippet from each document (this has some complications)

PR #561, opened 2016-11-13
[#560] FDA section for trials

nightsh:
So, this PR brings in the code *without* the text previews, which is to be added by #563 after the API gets to learn how to generate them (there's a PR there for it as well).

image

Fixes opentrials/opentrials#560


presenters/trial.js
```diff
@@ -58,6 +58,17 @@ function decorateDocuments(documents) {
...
 }
 
+function decorateFDADocuments(documents) {
+  const docs = _.filter(documents, (doc) => doc.source_id === 'fda');
+  _.map(docs, (doc) => {
+    if (doc.fda_approval) {
+      // eslint-disable-next-line no-param-reassign
+      doc.application_id = doc.fda_approval.fda_application.id;
+    }
+  });
```
vitorbaptista: It feels like you're using `map` here as if it was a `forEach`. The return value of a `map` is the initial array with the transformation of running its callback on each of its items. Here, you're instead altering the actual `doc` objects by adding an attribute to it. As they're objects (and not "primitives" like small integers), the `docs` array contain pointers to their variables, and not the variables themselves. This way, on `map`, you're changing the original models and not returning anything, so the result of it will be `[undefined, undefined, undefined, ...]` (as you're not returning anything).

That doesn't matter, though, because you're not returning the results of the map, but the original array (with items now changed).

You could change this code to either:

```javascript
const fdaDocs = _.filter(documents, { source_id: 'fda' })
return fdaDocs.map((doc) => {
  if (doc.fda_approval) {
    // eslint-disable-next-line no-param-reassign
    doc.application_id = doc.fda_approval.fda_application.id;
  }
  return doc;
});
```

Or even more "functionalish":

```javascript
return _.filter(documents, { source_id: 'fda' })
  .map((doc) => {
    if (doc.fda_approval) {
      // eslint-disable-next-line no-param-reassign
      doc.application_id = doc.fda_approval.fda_application.id;
    }
    return doc;
  });
```

nightsh: Good suggestions, I somehow overlooked that part of the code while being focused on the "make it pretty" side of things. Thanks for the suggestion, I'll push an update in a few minutes.

nightsh: Done 👍

nightsh: Well, I can't think of any javascriptish way to change the original value while iterating, other than working on a copy of the original array.

PR #567, opened 2016-11-17
Completions to the FDA documents section

nightsh:
This PR depends on opentrials/api#120, which should be merged before it.

image


nightsh:
Closing as we need to reconsider the design for displaying FDA documents.

Issue #594, opened 2016-12-06
Extract and expose last verified date from NCT


vitorbaptista:
For example, on https://clinicaltrials.gov/ct2/show/NCT00564096:

![transcranial magnetic stimulation tms in schizophrenia - full text view - clinicaltrials gov](https://cloud.githubusercontent.com/assets/76945/20916912/1911e258-bb97-11e6-8d10-88fb24d3c68c.png)

See https://github.com/opentrials/opentrials/issues/515 for more information on this issue.

georgiana-b:
In #515, me and @jessflem have determined that GSK also has the `last_validated` information. After a discussion you can read there, we concluded that:
1. `last_verified` should only be a property of `record` (not of `trial`)
2. the value we display for `last_verified` should be from the same `record` from which we took the `status` (i.e. primary record)

We already have in Processors [logic for determining which `record` is the primary source of data](https://github.com/opentrials/processors/blob/master/processors/base/writers/trial.py#L47) (i.e. from which `record` we take the values for a `trial`). However, that information is not stored anywhere so we can't decide which `last_verified` to display because we don't know from which `record` was the `status` taken.
This is problem also applies to trials that have multiple records from the same source [like this one that has 3 records from NCT](https://explorer.opentrials.net/trials/c4643df6-8bd4-11e6-a449-0242ac12000b).

Ways to go:
* store which `record` is primary
* don't display `last_verified`

@pwalsh, @jessflem What do you think?

pwalsh:
Store which record is primary

georgiana-b:
Closed via opentrials/processors#96, opentrials/api#132 and opentrials/opentrials#615

vitorbaptista:
@georgiana-b I asked this in another issue, but maybe it's best to keep the discussion here. Why don't we add `last_verified` to the `trials` table as well? I didn't understand why it's different than any other field we have.

PR #615, opened 2016-12-15
[#594] Show last_verified near status

georgiana-b:
opentrials/opentrials#594

Should be deployed after opentrials/api#132

last_verification_date

### DIFF
--- a/assets/styles/ot-scss/_lists.scss
+++ b/assets/styles/ot-scss/_lists.scss
@@ -156,7 +156,8 @@ aside ul {
 .overview-list {
 
   li {
-     padding: $small-spacing*0.75 $small-spacing*0.60;
+     padding-left: $small-spacing*0.60;
+     padding-right: $small-spacing*0.60;
     &.disabled {
       opacity: 0.8;
     }
--- a/assets/styles/ot-scss/_lists.scss
+++ b/assets/styles/ot-scss/_lists.scss
@@ -156,6 +156,7 @@ aside ul {
 .overview-list {
 
   li {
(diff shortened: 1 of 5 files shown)


assets/styles/ot-scss/_lists.scss
```diff
@@ -156,6 +156,7 @@ aside ul {
 .overview-list {
 
   li {
+     padding: $small-spacing*0.75 $small-spacing*0.60;
```
georgiana-b: This overwrites the right and left padding for overview-list so that all "boxes" (registration, gender, status, number of participants) can fit on one line.

views/trials-details.html
```diff
@@ -73,6 +73,19 @@
       </span>
       {{ add_discrepancy_icon_if_needed(trial, 'status') }}
     </li>
+    {% set primary_records = trial.records | selectattr("is_primary") | selectattr("last_verification_date") %}
```
nightsh: IMO this should belong to the trial presenter and have an associated test. See how the decorators are built in there (e.g. https://github.com/opentrials/opentrials/blob/master/presenters/trial.js#L61)

georgiana-b:
@nightsh I moved the logic in the presenter and added tests, please take a look.

georgiana-b:
@jessflem Is this what you had in mind for displaying `last_verified`?

jessflem:
@georgiana-b yes! excellent, thanks so much

nightsh:
We will only merge this after we run all the processors so we'd have `last_verified_date` in the response. This way we won't block the deployment of any other (potentially urgent) features.

Issue #573, opened 2016-11-29
Remove dead code and data


vitorbaptista:
There's some dead code on the Explorer that we're not using anymore. Review our code and remove them. We can always go back to our git history if we ended up needing something.

# Tasks

- [x] Remove `records.source_data` from the DB
- [ ] Remove records page (e.g. https://github.com/opentrials/opentrials/blob/master/views/trial-record.html and https://github.com/opentrials/opentrials/blob/2f8abde37faaa0bbc574a8a77bf8bb9072a2b788/handlers/trials.js#L24-L40)
- [ ] Remove publications handler/UI
- [ ] Review every view and handler to see if it's used. Remove the ones that aren't.

georgiana-b:
I noticed we have [some outdated documentation on database structure](https://github.com/opentrials/api/tree/master/docs).
@roll Do you remember what was the idea behind those docs? Isn't the up to date [schema described in swagger](https://api.opentrials.net/v1/docs/) enough?

roll:
@georgiana-b
It has been created even before swagger. Should be removed I suppose

georgiana-b:
Closed via opentrials/opentrials#617, opentrials/processors#97, opentrials/api#133

Issue #492, opened 2016-10-13
Publication search for academic journal articles and/or HRA lay summaries


benmeg:
Currently the advanced search option 'Publications'...'Trials with publications' appears to return both trials with linked academic journal articles and HRA summaries.

It would be useful to separate this out, so that a user looking just for trials with academic publications is not also given trials with just HRA lay summaries (or visa-versa - a user looking just for trials with HRA lay summaries, but not linked academic journal articles - admittedly a less plausible use case, but the point is to separate the two).

vitorbaptista:
@benmeg Maybe we could simply change the filter to return only publications as in journal articles, ignoring data from the HRA?

P.S.: We're considering HRA summaries as publications, but plan on separating them on #242

benmeg:
@vitorbaptista Yes, good idea - returning just linked academic journal articles would be a useful improvement

benmeg:
Once this is done, modify the 'Publications' tooltip (#315) to remove the second sentence i.e. "At present this also includes trials with lay summaries available from the UK Health Research Authority (HRA)."

PR #617, opened 2016-12-19
Remove unused code

georgiana-b:
opentrials/opentrials#573

I reviewed the other handlers & views but I didn't find any other that is unused.
I also removed the agents for `publications` and `records` because they were only used by the the handlers that were deleted.


handlers/publication.js
```diff
@@ -1,25 +0,0 @@
-'use strict';
```
nightsh: Add this back, see first comment and the comment on opentrials/opentrials#620.

routes/index.js
```diff
@@ -59,11 +54,6 @@ const routes = [
     config: require('../handlers/organisation'),
   },
   {
-    path: '/publications/{id}',
```
nightsh: Don't remove this route just yet (see opentrials/opentrials#620)

test/agents/publications.js
```diff
@@ -1,24 +0,0 @@
-'use strict';
```
nightsh: Add this back for now.

test/fixtures/index.js
```diff
@@ -5,10 +5,8 @@ const fixtures = {
   getOrganisation: require('./organisations'),
   getIntervention: require('./interventions'),
   getTrial: require('./trials').getTrial,
-  getRecord: require('./trials').getRecord,
   searchTrialsByEntity: require('./trials').searchTrialsByEntity,
   getLocation: require('./locations'),
-  getPublication: require('./publications'),
```
nightsh: We will keep this to be dealt with in opentrials/opentrials#620.

test/fixtures/publications.js
```diff
@@ -1,20 +0,0 @@
-const uuid = require('node-uuid');
```
nightsh: Keeping this too.

test/handlers/publications.js
```diff
@@ -1,63 +0,0 @@
-'use strict';
```
nightsh: Keep this for now.

views/publications-details.html
```diff
@@ -1,35 +0,0 @@
-{% extends "layouts/two-columns.html" %}
```
nightsh: Do not remove this yet.

views/trial-record.html
```diff
@@ -1,22 +0,0 @@
-{% extends "layouts/one-column.html" %}
```
nightsh: It's okay to remove it, as we don't link to it from anywhere yet. Just keep in mind we have this in git's history in case we'll need it in the future.

georgiana-b:
@nightsh  Ok, then let's just keep this open until #492 gets sorted.

georgiana-b:
@nightsh  I made separate a commit for `publications` so I propose we merge this if you're fine with it. If `publications` handlers & friends turn out to be needed in #492, that commit can be reverted.

Issue #503, opened 2016-10-18
Create alerts for data contributions


vitorbaptista:
They should send the notifications to our internal mailing list. We can use re:dash.

pwalsh:
add to body of email:

- the metadata
- a direct link to the file(s) on S3

vitorbaptista:
Blocked on https://github.com/opentrials/opentrials/issues/506

pwalsh:
Not blocked on #506 anymore - just needs to be done with Lambda / SNS etc. 

@vitorbaptista quite important as we have no visibility into contributions right now.

nightsh:
Here's a follow up of our latest discussion about this:

* do not implement the alerts in re:dash, but have them as an explorer feature
  * the new feature would use Mandrill as an email backend
  * this could probably use the `/lib/plugins` structure
* when a new data contribution is successfully added (promise is fulfilled), trigger a function using the alert functionality and:
  * send an email to OpenTrials Team <opentrials-team@okfn.org> (this could be an env var) with the new contribution details and relevant links

PR #623, opened 2016-12-22
[#503] Send email notification when a data contribution is submitted

georgiana-b:
opentrials/opentrials#503


package.json
```diff
@@ -60,6 +60,7 @@
     "bourbon-neat": "^1.7.4",
     "del": "^2.2.2",
     "dotenv": "^2.0.0",
+    "emailjs": "^1.0.8",
```
nightsh: Please update the shrinkwrap file too.

```bash
$ npm install
$ npm shrinkwrap
```

Then add it to git and commit.

**Note**: If shrinkwrap fails due to extraneous packages, run `npm prune` before to clean up your node_modules (things like this happen due to outdated shrinkwraps in dependencies and stuff like this).

test/lib/plugins/send-email-with-mandrill.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+const sendEmailWithMandrill = require('../../../lib/plugins').sendEmailWithMandrill;
+
+describe('sendEmailWithMandrill', () => {
+  describe('composeEmail', () => {
```
nightsh: Would be good to test the Markdown to HTML renderer too, if possible. Just to make sure it doesn't die due to invalid markdown in template (which is likely to happen when modifying the template).

views/email/test.md
```diff
@@ -0,0 +1 @@
+{{testID}}
```
nightsh: Can't we use the real email template for this?

test/lib/plugins/send-email-with-mandrill.js
georgiana-b: It's already tested on line 11.

views/email/test.md
georgiana-b: This test checks the validity of `composeEmail` in general not of data contribution notifications. That means it should pass whenever `composeEmail` does it's job instead of being dependent on a template that is subject to changes that may cause the test to break unnecessarily.

Issue #626, opened 2017-01-03
Ability to change data contribution category for curators


nightsh:
As a data curator, one has the ability to approve/reject user contributions with a message.

If the contribution is valid, therefore it should be approved, but its category is wrong, maybe we should allow the curators to change it pre-approval.

Just a thought, based mostly on the discussion in #558.

- [ ] Create categories drop down for each submission (or reuse the one in contribute-data page)
- [ ] revisit the [mapping document](https://docs.google.com/spreadsheets/d/1NvbNPC8cUrIg_uldmlB1B_1C_GmO03PrxrmEPeicjAc/edit#gid=0) and double check the remapped contributions
- [ ] TBD

vitorbaptista:
I'd say not only the category, but also the trial it's linked to. @georgiana-b I imagine this would be required to implement your solution on our data contributions pipeline?

nightsh:
FWIW, I volunteer to make it happen 😉

georgiana-b:
@vitorbaptista It is not required (I can make the processor without it) but I noticed curators currently try to make these corrections in the `curation_comments` so clearly these changes are needed. The curators should be able to change both `data_category_id` and `trial_id`. The sooner the better.

vitorbaptista:
Fixed

Issue #558, opened 2016-11-09
Revisions to 'Contribute Data' categories


jessflem:
@sarahcebm has reviewed the data schema for the 'Contribute data' categories.

Please remove the current list and replace with the revisions provided in [this document](https://drive.google.com/open?id=1oqJ95uVCzT-u9vy4SK8qK_U-osRnlzVUULf4Q3qzSwc). Upon submission all items should be tagged with the category and subcategory e.g 

Results - Journal article 
Results - Clinical Study report 
Study documents - Case report form 
Study documents - Grant application 

It should not be possible to submit only as 'Results'. The only exceptions are 'Registry entry' and 'Other'. 

Thanks!

benmeg:
N.B. When replacing the old category list we don't want to lose existing contributions re: their original categories - would some mapping be needed?

sarahcebm:
I'm happy to do some mapping if needed. 

Can you tell us how many docs have been submitted per existing category?

benmeg:
Paging @vitorbaptista

pwalsh:
@sarahcebm @jessflem 

I've added this to the [research requests board](https://github.com/opentrials/opentrials/projects/1), and I'll get one of the team to look at it in the week after we hand in the FDA app (so - first week of December).

Also adding @vitorbaptista and @nightsh to keep track.

pwalsh:
@vitorbaptista now assigning to you. can you please assign forward so it gets done this sprint.

pwalsh:
@nightsh this needs to be prioritized (we need it live by end of this coming sprint)

nightsh:
@jessflem @sarahcebm We would need some mapping, indeed. Not only for existing categories, but to create a migration file changing the values, so we need to cover all the cases (would be good to have it reversible and be able to roll back structural changes if needed).

You can add the mappings in the linked document, in the meantime I'll take care of the new structure and get the migration ready to be completed with the mentioned mappings.

pwalsh:
@nightsh See the question by @sarahcebm https://github.com/opentrials/opentrials/issues/558#issuecomment-259477413

And of course - don't bother migrating or mapping in cases were we don't even have data.

nightsh:
We currently have the following data contributed by users (count + category name):
```
19 | Publication (press release, article, blog post, etc.)
 6 | Clinical study report
 5 | Other
 4 | Patient information sheet
 4 | Trial protocols
 3 | Critical review
[...]

PR #629, opened 2017-01-05
New data contributions categories

nightsh:
Implement new categories structure in database, existing data and UI.

Some comments to be aware of before reviewing:

* all the publication related submissions have been merged under "Other" category and should be manually distributed to the many publication categories we have now (see #626)
* some other things are to be remapped to "Other", generally things that either have ambiguous destinations (category broken down to multiple pieces) or have neither contributions or remapping info available

This commit introduces two new migrations:

* one to add a `group` column to `data_categories`
* one to migrate existing contributions to the new categories (remapping)

They both have working `up` and `down` methods.

_P.S.: I also added a `npm run rollback` command to roll back migrations if we need. Just a minor thing that helps mostly in dev ;)_

screenshot_20170105_110320

opentrials/opentrials#558

### DIFF
--- a/migrations/20170103182459_data_categories_remapping.js
+++ b/migrations/20170103182459_data_categories_remapping.js
@@ -1,0 +1,144 @@
+'use strict';
+
+const _ = require('lodash');
+
+const newCategories = [
+  { name: 'Registry entry',
+    group: '',
+    replaces: ['Registry entry'] },
+  { name: 'Other',
+    group: '',
+    replaces: [
+      'Other',
+      'Publication (press release, article, blog post, etc.)',
+      'Trial webpage',
+      'Condition',
+      'Intervention',
+    ] },
+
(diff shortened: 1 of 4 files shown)


migrations/20170103182459_data_categories_remapping.js
```diff
@@ -0,0 +1,144 @@
...
+      );
+
+      const deleteAction = _.map(toDelete, (category) =>
+                                 knex('data_categories')
+                                 .where({ name: category.name, group: category.group })
+                                 .del()
+                                 .then()
+      );
+
+      return replaceAction && deleteAction;
```
georgiana-b: After rollback there are `data_contributions` that reference `data_categories` that no longer exist. Please make sure you also update foreign keys to `data_categories` in `down` or set them to null for categories that no longer exist.

Issue #622, opened 2016-12-21
Data contribution notifications don't dissapear


georgiana-b:
After submitting a data contribution, a notification appears on the screen prompting about success or failure. At the next submit, another notification appears, along with the previous one. See picture below: 

![screenshot_20161221_220907](https://cloud.githubusercontent.com/assets/8862002/21404950/5b91047e-c7cc-11e6-9f2b-7ecf123a88b6.png)

The most annoying part though is that they keep popping up on any page reload thereafter.

Environment: Firefox 50.0.2 for OpenSuse

benmeg:
Can confirm this using Chrome 55.0.2883.87 m on Windows 10

nightsh:
From the [yar documentation](https://github.com/hapijs/yar/blob/master/API.md):

> flash(type, message, isOverride) - stores volatile data - data that should be deleted once read. When given no arguments, it will return all of the flash messages and delete the originals. When given only a type, it will return all of the flash messages of that type and delete the originals. When given a type and a message, it will set or append that message to the given type. 'isOverride' used to indicate that the message provided should replace any existing value instead of being appended to it (defaults to false).

So we should probably look into (not) clearing up `yar.flash`. Or see if it's a library bug that was fixed in newer versions.

PR #630, opened 2017-01-06
[#622] Fix flash notifications not being discarded

nightsh:
There was an issue with the `yar` plugin being registered by Glue before the
extension function, thus leading to weird interferences with the session store.

Moved it to `lib/plugins/yar` and registered in `server.js` right after the
extensions.

Also styled the flash message so it doesn't look like it goes off screen.

image

Fixes opentrials/opentrials#622


nightsh:
Do not review yet, I am going to style the flash message a bit so it doesn't look like it goes off screen.

nightsh:
@vitorbaptista feel free to ping me if anything is unclear, I might have some relevant links in my browser history explaining the issue with `glue` and `yar` flash msgs.

nightsh:
I created a temp review app at https://ot-explorer-next-pr-630.herokuapp.com/

Issue #599, opened 2016-12-06
Display when a trial has received an exemption to publish results late


vitorbaptista:
For example, https://clinicaltrials.gov/ct2/show/NCT01982240 has this exemption (currently only ClinicalTrials.gov have this information). The data was extracted at https://github.com/opentrials/opentrials/issues/585.

This issue is to define how to display it and use the data already coming from the API.

georgiana-b:
Just a heads up for whoever picks up this task: I noticed we don't display  `completion_date` either.

PR #642, opened 2017-01-17
[#599] Display results_exemption_date in trial details

georgiana-b:
Fixes opentrials/opentrials#599

I displayed it in the `Documents` section just to be above `Results` since it's related info: 

results_exemption


vitorbaptista:
The wording makes it sound like the trial doesn't need to publish the results at all. Changing it to something like:

> This trial was exempted from publishing its results within 1 year of completion on 14/Sep/2001.

@opentrials/research Do you know of any place we can link that explains this type of exemption? I couldn't find any.

Also, please move the notice to the top of the main part of the page, just before the `Conditions` section.

Issue #650, opened 2017-01-24
Replace "type"  field in "documents" with "data_categories"


georgiana-b:
In #628 we decided to replace [the `type` field in `documents`](https://query.opentrials.net/queries/194) with  a `data_category`. That means we have to clone the `data_categories` table from `explorer` database into a `document/data_categories` table in `opentrials` API database and link that with `documents`.

georgiana-b:
@sarahcebm There is a data category called `Results - US Food and Drug Administration (FDA) document segment`.  What kind of FDA documents do you think fall under this category?

I'm asking because, as you might know, we already have some FDA documents on https://fda.opentrials.net :  `Aprroval Letters`, `Drug Label` etc. .Are these the kind of documents the above mentioned category refers to?

vitorbaptista:
Fixed

PR #664, opened 2017-01-30
[#650] Replace document type with document category

georgiana-b:
opentrials/opentrials#650

### DIFF
--- a/presenters/trial.js
+++ b/presenters/trial.js
@@ -4,15 +4,13 @@ const _ = require('lodash');
 
 function decorateDocuments(documents) {
   const nonFDADocuments = _.filter(documents, (doc) => doc.source_id !== 'fda');
-  const documentsDefaultGroup = _.remove(nonFDADocuments, (doc) =>
-    _.isUndefined(doc.document_category.group)
-  );
-  const groupedDocuments = _.groupBy(nonFDADocuments, (doc) =>
-    doc.document_category.group
-  );
-  if (documentsDefaultGroup.length > 0) {
-    groupedDocuments.Other = documentsDefaultGroup;
-  }
+  const groupedDocuments = _.groupBy(nonFDADocuments, (doc) => {
+    let groupName = doc.document_category.group;
+    if (_.isUndefined(groupName)) {
+      groupName = doc.document_category.name;
+    }
+    return groupName;
+  });
   return groupedDocuments;
 }
 
--- a/presenters/trial.js
+++ b/presenters/trial.js
@@ -3,59 +3,17 @@
 const _ = require('lodash');
 
 function decorateDocuments(documents) {
-  const documentTypesCategories = {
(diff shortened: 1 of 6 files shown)


georgiana-b:
@sarahcebm  This is how the trial documents look like grouped by `data_categories` `group` instead of the `display document type`. What do you think?

documents_with_contribute

sarahcebm:
Hi @georgiana-b  - is it possible to do a mock up with all data_categories group on display?  Perhaps with one place holder sub-category item available (green) and one not (red). Just to give an idea of when the entries become more densely populated with items.

georgiana-b:
@sarahcebm I know previously we displayed placeholders for non-available documents as in [this trial](https://explorer.opentrials.net/trials/8227fca0-e3bf-11e6-b65f-0242ac110002). However, because the sub-categories are much more numerous than the previously used `document types`,  I decided to display only the available documents.
Here is a version will a document for each group:

document_all_categories

The "Add new document" button is a link to the data contributions submission page and is displayed even when there are no documents. See example below: 

no_documents

If you think we should include placeholder sub-categories I think we need a selection of sub-categories that are more important and that we should display.

presenters/trial.js
```diff
@@ -3,59 +3,17 @@
...
-
-  return documentsByCategory;
+  const documentsDefaultGroup = _.remove(nonFDADocuments, (doc) =>
+    _.isUndefined(doc.document_category.group)
+  );
+  const groupedDocuments = _.groupBy(nonFDADocuments, (doc) =>
+    doc.document_category.group
+  );
+  if (documentsDefaultGroup.length > 0) {
+    groupedDocuments.Other = documentsDefaultGroup;
```
nightsh: This would overwrite the documents that already are in group "Other". Add to it instead of replacing.

test/fixtures/api/swagger.yaml
```diff
@@ -1017,7 +1017,7 @@ definitions:
       - id
       - name
       - url
-      - type
+      - document_category
```
nightsh: Can't we just leave it as `type`? It's more of a question, I'm not sure.

presenters/trial.js
georgiana-b: Reflecting your changes in #667, I changed the functionality to display `document_category.name` as group for those that don't have `group`.

test/fixtures/api/swagger.yaml
georgiana-b: As you've seen in  opentrials/api#143, `document_categories` is a table. We can't call that table `types` because: 
* `type` is used a lot throughout the database (for example [in `intervention`](https://github.com/opentrials/api/blob/master/api/models/intervention.js#L14)) where it represents a `string` *attribute* 
* it would mean that we keep all "types" in that table which is not true because we only keep document `type`

Issue #628, opened 2017-01-05
Process new data contributions


pwalsh:
@georgiana-b @nightsh this is quite urgent, especially the Ebola-related contributions.

georgiana-b:
@opentrials/research According to this task, we are going to transform data contributions into their corresponding entities in `opentrials` API database  ( `publications`, `documents` etc. ). In order to do this we need a mapping of data contribution categories to database entity and type.

I made [a spreadsheet here](https://docs.google.com/spreadsheets/d/1mNtceWbp_eoP9HrFkHiSF1AFvemJNK4Th4xS69qGOa4/edit?usp=sharing) with current data contribution categories. I mapped a few of them to the API database just for reference (maybe incorrectly, feel free to update). Please supply mapping for the other categories that should go into the API database.

For table `documents` we currently have a field `type` that can have only 3 values now: `csr_synopsis`, `other` and `results`. We will add other types requested in the mapping.

sarahcebm:
Hi @georgiana-b. Can I see the current listings for all entities so I can understand how it is currently mapped? The document field types you list don't match those on the trial entry e.g paperwork/regulatory documents/data. Also, is there a rationale for the current mappings that I can read?

georgiana-b:
@sarahcebm 

> Can I see the current listings for all entities so I can understand how it is currently mapped? 

Right now there is no mapping of data contributions to our `opentrials` database schema because we didn't process them so far. That's what we have to solve now.

> The document field types you list don't match those on the trial entry e.g paperwork/regulatory documents/data.

The more general categories you mentioned are [used to display the documents more conveniently](https://github.com/opentrials/opentrials/blob/master/presenters/trial.js#L6). We could get rid of those in the future since they require yet another mapping to maintain.
I was referring to the [types of documents we have in the database](https://query.opentrials.net/queries/194). For convenience, here is a list with all document types we allow at the moment and the category used for displaying them:

| database document type | display document type |
|----|----|
| blank_consent_form | Paperwork |
| patient_information_sheet | Paperwork |
| blank_case_report_form | Paperwork|
| csr | Regulatory documents |
| csr_synopsis | Regulatory documents |
| epar_segment | Regulatory documents |
| results | Data |

[...]

PR #677, opened 2017-02-09
[#628] Add document_id to contributions

georgiana-b:
I also changed the empty groups to `null` instead of empty string.
opentrials/opentrials#628

### DIFF
--- a/migrations/20170131152728_add_document_id_to_data_contributions.js
+++ b/migrations/20170131152728_add_document_id_to_data_contributions.js
@@ -1,0 +1,11 @@
+'use strict';
+
+exports.up = (knex) => (
+  knex.schema.table('data_contributions', (table) =>
+    table.uuid('document_id')
+  )
+);
+
+exports.down = (knex) => (
+  knex.schema.table('data_contributions').dropColumn('document_id')
+);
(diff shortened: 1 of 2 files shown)


nightsh:
@georgiana-b who's writing in the new `document_id` field?

georgiana-b:
@nightsh The `document_id` field will be populated by `data_contributions` processor if a `data_contribution` is successfully turned into a `document`. Here [is the relevant code](https://github.com/opentrials/processors/blob/f9a38b026e003b818b22cc79e8aeacc0c9462450/processors/data_contributions/processor.py#L76) from the PR containing `data_contributions` processor. The PR is still unmerged and will remain that way until #685 gets sorted but merging this can be done safely.

migrations/20170131152728_add_document_id_to_data_contributions.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
+
+exports.up = (knex) => (
+  knex.schema.table('data_contributions', (table) =>
+    table.uuid('document_id')
+  )
+);
+
+exports.down = (knex) => (
+  knex.schema.table('data_contributions').dropColumn('document_id')
```
nightsh: This happens on `migrate:rollback`:

```
Using environment: development
Knex:warning - migrations failed with error: knex.schema.table(...).dropColumn is not a function
```

PR #681, opened 2017-02-13
[#626][#657][#692] Allow updating data contribution category and trial ID

### LINKED ISSUE #626: Ability to change data contribution category for curators

### LINKED ISSUE #657: Link to contributions curation page from notification email
Quoting @bengoldacre:

> also, better if it could link to the page of submissions?

Task here would be to add a link to the curation page in the notification email.

### LINKED ISSUE #692: E-mail with data contribution URLs breaks with URLs containing whitespaces
My guess is that we're not encoding the URLs.

nightsh:
Added a link to data curation page in the data contribution notification email (fixes #657).

Encode URLs in the data contribution notification email (fixes #692).

New fields for data contributions curation page (fixes #626):

* `data_category_id` – preselected the current category, allows updating
* `trial_id` – prefilled with current trial ID, allows updating / clearing

There are also some cosmetic changes to the curation page:

* removed "Approved?", "Comments" and "Curator's comments" columns
* color coded table rows according to approval status
* moved contributor's and curator's comments on a row below the submission, to
  clear up some space in the UI
* renamed "Approve" and "Reject" buttons on approved, respectively rejected
  submissions (now it just says "Save", which is less misleading)

image


handlers/admin/data-contributions.js
```diff
@@ -16,23 +19,25 @@ function editDataContribution(request, reply) {
...
+    .then((categories) => new DataContribution()
+          .query('orderBy', 'created_at', 'desc')
+          .fetchAll({ withRelated: DataContribution.relatedModels })
+          .then((dataContributions) => {
+            reply.view('admin/data-contributions', {
+              title: 'Data contributions',
+              dataContributions: dataContributions.toJSON(),
+              categories: categories.toJSON(),
+            });
+          }));
```
vitorbaptista: I think using `Promise.join()` can improve the readability here, like:

```javascript
return Promise.join(
  new DataCategory().orderBy('name').fetchAll(),
  new DataContribution()
    .query('orderBy', 'created_at', 'desc')
    .fetchAll({ withRelated: DataContributions.relatedModels }),
  (categories, dataContributions) => {
    reply.view(/* ... */)
  })
```

You could even go a step further and extract the `.fetchAll()` calls as constants like:

```javascript
const fetchDataCategories = new DataCategory()
  .orderBy('name')
  .fetchAll();
const fetchDataContributions = new DataContribution()
  .query('orderBy', 'created_at', 'desc')
  .fetchAll({ withRelated: DataContributions.relatedModels });

return Promise.join(
  fetchDataCategories,
  fetchDataContributions,
  (categories, dataContributions) => {
    reply.view(/* ... */)
  })
```

test/handlers/admin/data-contributions.js
```diff
@@ -81,12 +81,19 @@ describe('admin data contributions handler', () => {
...
 
+      factory.create('dataCategory')
+        .then((dataCategory) => {
+          payload = {
+            approved: true,
+            trial_id: '11111111-1111-1111-1111-111111111111',
+            data_category_id: dataCategory.id,
+            curation_comments: 'My comments',
+          };
+        });
```
vitorbaptista: There's a race condition here. As these promises aren't part of the promises we return, there's no guarantee that they would be executed in order. If they take too long, the test can even finish before they finish executing. To solve this, you must move it to the returned promises.

You could add it between the `.then((dataContribution) => (dataContributionId = dataContribution.attributes.id))` and the `server.inject` lines, simply creating the new data category and changing the `payload` with its ID.

test/handlers/admin/data-contributions.js
```diff
@@ -81,12 +81,19 @@ describe('admin data contributions handler', () => {
     });
 
     it('updates the DataContribution and redirects to /admin/data-contributions', () => {
-      const payload = {
-        approved: true,
-        curation_comments: 'My comments',
-      };
```
vitorbaptista: No need to change it from a `const` to a `let`, as you can change constant dict's internal values. You can just initiate it with the hardcoded values (i.e. `approved`, `trial_id` and `curation_comments`) and then add the `data_category_id` when you create a new data category.

views/admin/data-contributions.html
```diff
@@ -67,20 +65,54 @@
...
       </td>
-      <td>
-        {{ dataContribution.comments }}
-      </td>
 
       <td>
         <form action="/admin/data-contributions/{{ dataContribution.id }}" method="POST">
+          <select id="data_category_id" name="data_category_id">
+            <option value="" disabled selected>Select a new category (optional)</option>
+            {% for group, categories in  categories | sort(false, false, 'group') | groupby("group") %}
```
vitorbaptista: Double whitespace here and mixing up single and double quotes

views/admin/data-contributions.html
```diff
@@ -5,27 +5,25 @@
...
-        {% else %}
-          No
-        {% endif %}
-      </td>
+      {% if dataContribution.approved %}
+        {% set status = "approve" %}
+      {% elif dataContribution.approved === false %}
+        {% set status = "reject" %}
+      {% else %}
+        {% set status = "unmoderated" %}
```
vitorbaptista: As the class `unmoderated` isn't used for anything, you could simply set `status = ""`

vitorbaptista:
@nightsh I added a few comments on the code, mostly aesthetic.

In regards of the design, I think the buttons are confusing. It feels a bit overkill, but I think having 3 buttons (Save, Approve, Reject) at all times is the best solution.

I also don't like that the place to change the trial ID is separate from where we display it, but I understand you've done that so we can keep it as a link. Good enough for now IMHO.

assets/styles/ot-scss/_data-curation.scss
```diff
@@ -11,6 +11,20 @@
   tr {
     vertical-align: top;
     overflow: hidden;
+    border-bottom: 0;
+
+    &.unmoderated {
+      background-color: transparent;
+    }
```
nightsh: Fair enough, I'll remove it.

handlers/admin/data-contributions.js
```diff
@@ -1,13 +1,16 @@
...
 const DataContribution = require('../../models/data-contribution');
 
 
 function editDataContribution(request, reply) {
   const data = {
     approved: request.payload.approved,
-    curation_comments: request.payload.curation_comments,
+    trial_id: request.payload.trial_id || null,
+    data_category_id: request.payload.data_category_id,
+    curation_comments: request.payload.curation_comments || null,
```
nightsh: Yeah, nunjucks is not the most friendly system with undefined and null values. It will attempt to cast them into a literal string quite often. I don't remember exactly what was the issue here, but I'd dare guessing that if it were `undefined` then nunjucks would fail its `{% if curation_comments %}` blocks due to casting the `undefined` into `'undefined'`. Honestly though, I don't remember exactly, want me to check?

handlers/admin/data-contributions.js
nightsh: Oh, can you do that?! :hushed: 

You, sir, won the JavaScript award today! Thanks so much :)

handlers/admin/data-contributions.js
vitorbaptista: If by check you mean "remove the `|| null` and see what breaks", taking a couple minutes, then yes. This type of weird-ish code has a tendency of living on in my experience, as people don't want to change what they don't understand :smile:

test/handlers/admin/data-contributions.js
nightsh: I think the linter was crying about it. Will recheck and tell.

test/handlers/admin/data-contributions.js
nightsh: Ah, I see. This sounds a bit familiar, indeed, I guess I've seen some behaviours similar with the one you're describing, in other tests. Good catch!

views/admin/data-contributions.html
nightsh: OK, that was mostly a "just-in-case" half-feature, in case we'd want to highlight the unmoderated things somehow. But I figured that'd create quite a rainbowy look to the page, and since we're not in the 70s anymore, I left it in a resting position. I'll just get rid of it :)

views/admin/data-contributions.html
nightsh: Ha ha, you're better than `eslint` sometimes :laughing: 
Thx

nightsh:
@vitorbaptista I processed your suggestions, ready for another look.

And hey, this was one of the most useful code reviews ever, I'm still thankful for the `Promise.join()` thingy. Made me to take a closer look at `bluebird` documentation. Kudos! :bowing_man:

test/handlers/admin/data-contributions.js
```diff
@@ -83,10 +83,17 @@ describe('admin data contributions handler', () => {
...
+        trial_id: '11111111-1111-1111-1111-111111111111',
+        data_category_id: null,
         curation_comments: 'My comments',
       };
       let dataContributionId;
 
+      factory.create('dataCategory')
+        .then((dataCategory) => {
+          payload.data_category_id = dataCategory.id;
+        });
```
vitorbaptista: There's still the same race condition. This promise needs to be part of the chain that's returned, like:

```javascript
return factory.create('dataContribution')
  .then(() => ())
  .then(() => factory.create('dataCategory'))
  .then((dataCategory) => (payload.data_category_id = dataCategory.attributes.id))
  // ...
```

It's also possible to refactor this using the `Promise.join` like:

```javascript
return Promise.join(
    factory.create('dataContribution'),
    factory.create('dataCategory'),
    (dataContribution, dataCategory) => {
      dataContributionId = dataContribution.attributes.id;
      payload.data_category_id = dataCategory.attributes.id;
    }
  )
  .then(() => server.inject({}))
```

handlers/admin/data-contributions.js
```diff
@@ -16,23 +20,30 @@ function editDataContribution(request, reply) {
       request.yar.flash('success', 'Data Contribution updated successfully.');
       return reply.redirect('/admin/data-contributions');
     })
-    .catch(() => {
-      // FIXME: Log error
+    .catch((error) => {
+      console.error(error.detail); // eslint-disable-line no-console
```
vitorbaptista: I'm not sure what's in `error.detail`, but we log as `console.error(error)` in the API and it returns the traceback. Feel free to keep whatever you prefer, just saying in the unlikely case you haven't tried it as well.

test/handlers/admin/data-contributions.js
nightsh: Can we discard this for now? It has been removed by #691 just recently, so it's going to die anyway.

handlers/admin/data-contributions.js
nightsh: Can we have something like... a log file? I'm generally not happy having `console.whatever()` statements in the code, due to their usual `STDOUT`/`STDERR` output. While that's not inherently bad, it might be harder to observe in our case.
Also, would it make sense for us to create a "standardise logging throughout the whole app" issue? So we'd send things to a global logger that pipes the output through to various places.

vitorbaptista: Those logs from console are already sent to our Papertrail account, so all's good. In the future, I'd like to enable Sentry, as I find it more useful than just the logs.

pwalsh: yes, just to iterate on what @vitorbaptista said - STDOUT **is** our "standardised logging" interface, @nightsh . As we use docker, our app environments are stateless, and we don't rely on files or data co-located with the app. So for logs, we pipe the stream to papertrail.

vitorbaptista:
Closing this as all of its commits are part of https://github.com/opentrials/opentrials/pull/691 as well. Will merge the other PR instead.

PR #691, opened 2017-02-23
[#685] Consume document categories form API instead of local models

### LINKED ISSUE #685: Remove data contribution categories from explorer database structure
As of now, we ended up needing data contributions categories in both API and Explorer database. This creates all kinds of syncing issues between the two databases, which would be easily avoidable by storing them in just one place.

Following the same rationale that got us into crafting #677 (add API `document_id` to explorer contributions), we need to:

- [x] create table in API that will keep categories
- [x] populate the above mentioned table with existing categories
- [x] remove `document_categories` table and related logic from API
    - [x] Or recycle `document_categories` 
- [x] create API endpoint for getting categories list
- [x] create API endpoint for getting a category by id
- [x] adjust Swagger to reflect changes
- [x] write API tests
- [x] remove `data_contribution_categories` table and its related constraints from explorer
- [x] populate the categories list using an API call
- [x] adjust tests and factory (probably also write an e2e test for contributions)

This will also involve some changes in `opentrials/processors` since we will no longer need to sync the `data_contribution_categories` table with `document_categories`:
- [x] remove  the logic related to syncing especially the ability to create and update `document_categories`
- [x] refactor the way we refer to `document_categories`: create a single point of reference to `document_categories` by `id`(mapping) and use the alias declared in the mapping throughout the processors
- [x] update schemas  for API and explorer test databases

nightsh:
Depends on #681 to be merged first (its commits are rebased into this one).
Fixes #685 

This changes the way we're addressing data contributions categories by
requesting them from the OT API instead of keeping a local database table and
associated model. This was done to ease automated processing of the contributed
data and avoid having duplicate tables between Explorer and API (see #685).

New migration to:

* remove data categories
* rename `data_category_id` to `document_category_id`

Document categories integration:

* agent instead of local model
* replaced model instance with agent promise throughout the code
* adjusted all the tests, the factory and fixtures data
* updated `swagger.yaml` fixture

On the views:

* new filter `getAttributeById` to pluck an attribute of an object from an array
  of `id` identified objects.
  For example `{{ categories | getAttributeById(value, 'name')}}` will get the
  value for the `name` attribute of the list element having `id` equal to
  `value`.
* use the new filter in contributions curation page to avoid unnecessary API
  calls for getting the name of the assigned category for each listed item


views/admin/data-contributions.html
```diff
@@ -2,70 +2,76 @@
...
-  <tbody>
-    {% for dataContribution in dataContributions %}
-      {% if dataContribution.approved %}
-        {% set status = "approve" %}
-      {% elif dataContribution.approved === false %}
-        {% set status = "reject" %}
-      {% else %}
-        {% set status = "" %}
-      {% endif %}
+<section class="data-curation">
```
nightsh: Woohoo, going semantic! \o/

Thanks for doing that. In the same time, I'm screaming internally because I had this itch for weeks at least :sweat_smile:

nightsh:
@vitorbaptista, is it reviewable?
@georgiana-b, you're the only one left who hasn't touched this branch (at least, yet). So you get to play with it in the end :)

PR #706, opened 2017-03-18
[#652] Pressing ENTER on autocomplete fields submits form

vitorbaptista:
Fixes opentrials/opentrials#652

### DIFF
--- a/assets/js/index.js
+++ b/assets/js/index.js
@@ -3,8 +3,16 @@
 
   function setupSelect2For(name) {
     const perPage = 10;
+    const input = $('select[name="' + name + '"]');
+    const form = input.parents('form');
 
-    $('select[name="' + name + '"]').select2({
+    function _submitFormIfPressedEnter(e) {
+      if (e.keyCode === 13) {
+        form.submit();
+      }
+    }
+
+    input.select2({
       ajax: {
         url: OPENTRIALS_API_URL + '/search/autocomplete/' + name,
         dataType: 'json',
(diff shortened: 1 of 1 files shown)


assets/js/index.js
```diff
@@ -3,8 +3,16 @@
 
   function setupSelect2For(name) {
     const perPage = 10;
+    const input = $('select[name="' + name + '"]');
+    const form = input.parents('form');
 
-    $('select[name="' + name + '"]').select2({
+    function _submitFormIfPressedEnter(e) {
+      if (e.keyCode === 13) {
```
nightsh: I would also stick this in here:
`input.prop("disabled", true);`

assets/js/index.js
```diff
@@ -35,6 +43,10 @@
         },
       },
       tags: true,
+    }).on('select2:select', function () {
+      form.off('keyup', '.select2-selection', _submitFormIfPressedEnter);
+    }).on('select2:opening', function () {
```
nightsh: I tried binding it to other events, to no success. Ideally, we won't have to wait for s2 to open so we could `submit()` the form, but if there's no clear way to do it otherwise, then I'm okay with this. I'd just disable the input upon submitting, imho it improves a bit the feedback you get from pressing enter.

assets/js/index.js
vitorbaptista: I tried with it, but felt it made the "blinking" worse, as it adds another blink where the input becomes grey.

It's stupid that there's no easier way of allowing ENTER to submit the form, which means we have to live with these hacks and the flash of new autocomplete suggestions appearing between pressing ENTER and the form being submitted.

nightsh: well, we could hack it even more and sticking a `hide()` in there :skull_and_crossbones: 

...or we could leave it this way.

nightsh:
Hey @vitorbaptista I'd say let's move this forward. Unless you have any updates to the PR, so letting you push the buttons.

Issue #651, opened 2017-01-24
Disable autocomplete feature for everything except "locations"


vitorbaptista:
The autocomplete results are messy because the underlying data is messy, so they confuse more than help. The exception here is for location information, as we're normalizing it on https://github.com/opentrials/processors/pull/102.

# Tasks

- [x] Convert conditions, interventions, people and organisations filters to regular text inputs (rename their label to singular)
- [x] Change the Swagger file to allow autocomplete only on `location`
- [x] Change the indexers to index only locations

nightsh:
We also need to adjust the search a bit. If we don't autocomplete anymore, this will lead to a world of typos, so fuzzy search should be default for all of them (I just did a search for condition "enrgy intake" and got me nothing, while "energy intake" has results).

vitorbaptista:
Sure, but we can handle that issue separately, as it involves the whole search infrastructure

nightsh:
Opened an issue for it, so we can now focus on this one.

georgiana-b:
Closed via opentrials/api#154 and opentrials/opentrials#719.

PR #719, opened 2017-04-04
[#651] Disable autocomplete for all fields except locations

vitorbaptista:
As our fields are still messy, not normalized, the autocomplete fields are more
confusing than helpful. The only exception is the locations, which are
relatively well normalized.

opentrials/opentrials#651


views/partials/search-form-fields.html
```diff
@@ -24,10 +24,34 @@
...
-  {{ createSelectMultiple('condition', 'Conditions', 'Disease, disorder, syndrome, illness or injury that is being studied.') }}
-  {{ createSelectMultiple('intervention', 'Interventions', 'Process or action that is the focus of a clinical study. This can include giving participants drugs, medical devices, procedures, vaccines, and other products that are either investigational or already available. Interventions can also include noninvasive approaches such as surveys, education, and interviews.') }}
-  {{ createSelectMultiple('person', 'People') }}
-  {{ createSelectMultiple('organisation', 'Organisations') }}
+  <div class="filter">
+    <label
+        for="condition"
+        class="hastip"
+        title="Disease, disorder, syndrome, illness or injury that is being studied."
+    >Condition</label>
```
nightsh: Since we're here and we're disabling autocomplete, should we also rename this to "Medical Condition" or something more relevant? We're kinda taking away part of its context, so we need to make it more explicit now.

views/partials/search-form-fields.html
```diff
@@ -24,10 +24,34 @@
...
+    <label
+        for="intervention"
+        class="hastip"
+        title="Process or action that is the focus of a clinical study. This can include giving participants drugs, medical devices, procedures, vaccines, and other products that are either investigational or already available. Interventions can also include noninvasive approaches such as surveys, education, and interviews."
+    >Intervention</label>
+    <input title="intervention" name="intervention" value="{{ query.intervention }}" />
+  </div>
+
+  <div class="filter">
+    <label for="person">People</label>
```
nightsh: Same as for conditions, this is a good opportunity to clear up some confusing terms.

vitorbaptista:
@nightsh As the name changes is a different issue and require input from @opentrials/research, I'm merging this :v:

Issue #715, opened 2017-04-03
Hootsuite shortened links give error


benmeg:
By default Hootsuite shortened links are [suffixed with ?platform=hootsuite](https://forum.hootsuite.com/discussion/6614888/how-to-stop-appending-platform-hootsuite-to-url)

Seen via this tweet: https://twitter.com/CochraneUK/status/848819881432805376

Short link: http://ow.ly/maQk30auXom

Expanded link: https://fda.opentrials.net/search?platform=hootsuite

Error displayed: Error 400: Bad Request - "platform" is not allowed

Also present on main explorer: https://explorer.opentrials.net/?platform=hootsuite

PR #733, opened 2017-04-06
[#715] Strip unknown query params when in production

vitorbaptista:
Some link shorteners, like Hootsuite, add query params to its links, which then
breaks our Joi query params validations (as we're receiving an unknown param).
Instead of throwing an error, we now simply ignore and remove it from the params
passed to our handlers. This can cause some confusion as on why the handler
isn't receiving a param, but is a better experience for our users.

Fixes opentrials/opentrials#715


routes/index.js
```diff
@@ -105,4 +107,21 @@ const routes = [
...
-module.exports = routes;
+function stripUnknownParamsOnProduction(route) {
+  const inProduction = (process.env.NODE_ENV === 'production');
+  const result = _.cloneDeep(route);
+
+  if (inProduction && result.config && result.config.validate) {
+    const options = result.config.validate.options || {};
+
+    if (!Object.prototype.hasOwnProperty.call(options, 'stripUnknown')) {
+      options.stripUnknown = true;
```
nightsh: This is so cool, I didn't know we have this option! Great job!

test/routes/index.js
```diff
@@ -0,0 +1,46 @@
...
+  describe('stripUnknown', () => {
+    const originalEnv = Object.assign({}, process.env);
+
+    beforeEach(() => {
+      delete require.cache[require.resolve('../../routes')];
+    });
+
+    afterEach(() => {
+      delete require.cache[require.resolve('../../routes')];
+      process.env = originalEnv;
```
nightsh: Nice touch, I actually came here to suggest it and :boom: you had it in already :wink:

routes/index.js
```diff
@@ -105,4 +107,21 @@ const routes = [
   },
 ];
 
-module.exports = routes;
+function stripUnknownParamsOnProduction(route) {
+  const inProduction = (process.env.NODE_ENV === 'production');
+  const result = _.cloneDeep(route);
```
nightsh: Why not work directly on `route`, besides good Node habits? I'm just curious if you observed any weird side effects.

test/routes/index.js
```diff
@@ -0,0 +1,46 @@
...
+    });
+
+    describe('not on production', () => {
+      it('is undefined', () => {
+        process.env.NODE_ENV = 'foobar';
+        const routes = require('../../routes');
+
+        routes.forEach((route) => {
+          if (route.config && route.config.validate && route.config.validate.options) {
+            should(route.config.validate.options.stripUnknown).be.undefined();
```
nightsh: This is likely to fail the moment we intentionally add `stripUnknown` to any of the routes. Is that acceptable?

routes/index.js
vitorbaptista: There was some issue with the tests. For some reason, when invalidating the require cache and re-loading the file, we still got the old routes. I didn't enter this rabbit hole :P

test/routes/index.js
vitorbaptista: Good point. However, it's too hard to filter out the routes that add `stripUnknown` in their routes files, so I don't think it's worth it. Specially considering that in the worst case the tests will simply fail, so we'll know there's a problem.

routes/index.js
vitorbaptista: Me neither. It was a good surprise :)

Issue #656, opened 2017-01-25
Create trial entity for unregistered trials from PubMed


vitorbaptista:
Unregistered trials are defined on #655. Populate the fields:

* Title -> publication's title
* Summary -> publication's abstract
* Publications -> the publication itself

They should have some UI element to differentiate them in the search results from the other registered trials, and potentially remove things that they'll never have like `conditions` and `interventions`

georgiana-b:
@vitorbaptista I have some questions regarding unregistered trials.

Context: Currently [we extract `trial` identifiers from `publications` `title` and `abstract` attributes](https://github.com/opentrials/processors/blob/master/processors/pubmed_linker/processor.py#L23). This is a very loose linking since it doesn't guarantee that a `trial` reference means the `publication` is a byproduct of that `trial` (people could reference a `trial` in the `abstract` for all sorts of reasons). We have a way to make this process more specific that I described in #682.

Having these said, here are the questions:
1. We consider a `publication` to be an unregistered `trial`:
   a. if it doesn't have any references to `trials` in our database
   b. if it doesn't have identifiers that suggest it belongs to a `trial` registered in other sources (i.e. no values in `DataBankList`(see #682)

2. If an "unregistered" `trial` has been created from a `publication`, do we update the attributes of the `trial` every time the `publication` is updated?

3. If a `publication` is updated with identifiers (e.g. NCT ID is added to `DataBankList`) we can conclude that the unregistered `trial` created from it is no longer unregistered and delete it? 
Even if we don't have a `trial` with that NCT ID in our database?

vitorbaptista:
> 1. We consider a `publication` to be an unregistered `trial`:
>    a. if it doesn't have any references to `trials` in our database
>    b. if it doesn't have identifiers that suggest it belongs to a `trial` registered in other sources (i.e. no values in `DataBankList`(see #682)

Sounds sensible

> 2. If an "unregistered" `trial` has been created from a `publication`, do we update the attributes of the `trial` every time the `publication` is updated?

Yes

> 3. If a `publication` is updated with identifiers (e.g. NCT ID is added to `DataBankList`) we can conclude that the unregistered `trial` created from it is no longer unregistered and delete it?  
> Even if we don't have a `trial` with that NCT ID in our database?

[...]

PR #738, opened 2017-04-09
[#656] Optimize search results display for unregistered trials

georgiana-b:
opentrials/opentrials#656

I changed the search results page to be more unregistered trials friendly. This is how it looks now. The last 2 trials are unregstered:

unregistered_trials_pr


views/partials/trial-search-result.html
```diff
@@ -2,21 +2,29 @@
...
+    {% set status = trial.status or unknownValue %}
     <dt title="Status" class="progress"><span>Status:</span></dt>
     <dd>{{ status | underscoresToCapitalized }}</dd>
 
-    {% set recruitment_status = trial.recruitment_status or 'unknown' %}
+    {% set recruitment_status = trial.recruitment_status or unknownValue %}
     <dt title="Recruitment status" class="recruitment"><span>Recruitment status:</span></dt>
     <dd>{{ recruitment_status | underscoresToCapitalized }}</dd>
 
+    {% set source_name = trial.source.name or (unknownValue | underscoresToCapitalized) %}
```
vitorbaptista: Feels like it should be `(trial.source.name | underscoresToCapitalized) or unknownValue`, no? (Considering `underscoresToCapitalized` returns `undefined` on `undefined` values

views/partials/trial-search-result.html
```diff
@@ -30,7 +38,9 @@ <h1 class="title"><a href="/trials/{{ trial.id }}">{{ trial.public_title }}</a><
     <dd>
       <ul>
         {% for condition in trial.conditions %}
-        <li>{{ condition.name }}</li>
+          <li>{{ condition.name }}</li>
+        {% else %}
+          <p>{{ unknownValue | underscoresToCapitalized }}</p>
```
vitorbaptista: I think HTML like `<ul><p></p></ul>` is invalid, so we either need to check if `trial.conditions` exist and either create the `<ul>` or not (no need for a `<p>` here, as we're already inside the `<dd>`), or add the unknown value inside a `<li>`. I prefer the former.

views/partials/trial-search-result.html
```diff
@@ -40,6 +50,8 @@ <h1 class="title"><a href="/trials/{{ trial.id }}">{{ trial.public_title }}</a><
       <ul>
       {% for intervention in trial.interventions %}
         <li>{{ intervention.name }}</li>
+      {% else %}
+        <p>{{ unknownValue | underscoresToCapitalized }}</p>
```
vitorbaptista: Same issue as the above comment.

Issue #729, opened 2017-04-06
Cannot authenticate user


nightsh:
Trying to login with either Google or Facebook returns 400 Bad Request with the message:

> Auth failed: Invalid setting - isSecure must be set to false for non-https server

This was reproduced in production.

nightsh:
![image](https://cloud.githubusercontent.com/assets/1579997/24744878/30757a16-1abc-11e7-9293-7e7180faf1c9.png)

PR #743, opened 2017-04-10
[#729] Fix bug where couldn't authenticate on HTTPS

vitorbaptista:
Fixes opentrials/opentrials#729


vitorbaptista:
It's hard to check that this PR works, as it depends on external systems. However, you can check the deployed instance on https://ot-explorer-next-pr-743.herokuapp.com/. If you try to log in, the error is on Google (or Facebook)'s side, which means that the `isSecure` problem was fixed.

PR #745, opened 2017-04-11
 [#708] Add filter by source

### LINKED ISSUE #708: Add filter by source
Add a filter that allows people picking which sources they'd want to search for. For the UI, you can use a similar implementation as the autocomplete boxes, but that doesn't allow writing new values (i.e. can only use valid names).

There's an endpoint on https://api.opentrials.net/v1/sources to get all of our sources, but be aware that not all sources are trial sources. We have things like `fda`, which shouldn't appear in this list. We might need to change our API responses to develop this feature. Please analyse the issue and write a proposal on the solution before developing it.

nightsh:
This adds a new filter on the search sidebar to look for trials having a `source_id` corresponding to any of the variants chosen. By default, all available sources are selected.

Sources to choose from will have to have `type` equal to `register` or `journal`.

Closes #708


handlers/search.js
```diff
@@ -91,28 +95,36 @@ function searchPage(request, reply) {
...
+        pagination,
+        sourcesList,
+        advancedSearchIsVisible: advancedSearch,
+        trials: _trials,
+      });
     });
-  }).catch((err) => (
+  }
+  )
+  .catch((err) => (
```
vitorbaptista: If I remember correctly, we've been styling the code as:

```
foo()
  .then(() => {
    // ...
  })
  .catch((err) => {
    // ...
  })
```

handlers/search.js
```diff
@@ -91,28 +95,36 @@ function searchPage(request, reply) {
...
-    const pagination = generatePaginationLinks(
-      request.url, currentPage,
-      perPage, maxPages,
-      _trials.total_count
-    );
-    const queryDescription = queryStr || 'clinical trial data';
-    const title = `Search results for ${queryDescription} in OpenTrials`;
-    const description = `OpenTrials has ${_trials.total_count} matches for
+  sources.list().then((_sources) => {
+    trials.search(queryStr, page, perPage, filters).then((_trials) => {
```
vitorbaptista: You have a bug here. As you're not returning the inner promise, JavaScript doesn't guarantee that it'll be ran before the rest of the code. You need to return it, so all promises are part of the same chain and JS guarantees that they'll run in sequence. The styling that could be improved as well. I think the best would be:

```javascript
Promise.join(
  sources.list(),
  trials.search(queryStr, page, perPage, filters),
  (_sources, _trials) => {
    // ...
  }
).catch((err) => {
  // ...
})
```

handlers/search.js
```diff
@@ -91,28 +95,36 @@ function searchPage(request, reply) {
...
-      title,
-      description,
-      query,
-      currentPage,
-      pagination,
-      advancedSearchIsVisible: advancedSearch,
-      trials: _trials,
+      const sourcesList = _sources.filter((s) => (
+        ['register', 'journal'].indexOf(s.type) > -1)
+      );
```
vitorbaptista: The formatting here is quite confusing. I think this could either be in a single line, or move the closing parenthesis in line 111 to 112, so:

```javascript
const sourcesList = _sources.filter((s) => (
  ['register', 'journal'].indexOf(s.type) !== -1
));
```

(I also prefer checking `!== -1` than `> -1`, but there's no argument for/against each style, as they both have the same result.)

views/partials/search-form-fields.html
```diff
@@ -159,6 +159,23 @@
   </div>
 
   <div class="filter">
+    <label for="source"{% if tip %} class="hastip" title="Trial source"{% endif %}>Sources</label>
```
vitorbaptista: No need to check `{% if tip %}` here.

assets/js/index.js
```diff
@@ -88,5 +88,6 @@
     });
 
     setupSelect2For('location');
+    $('select[name="source"]').select2();
```
vitorbaptista: I'm not sure (and haven't checked the API docs), but for these sources we don't want to allow the user to search for a source name that doesn't exist in our data. This is different from the `location` select, where the user can add anything. Is this the case here?

handlers/search.js
vitorbaptista: You could also extract the inner function `(_sources, _trials) => {}` to a private function. The complicator is that it requires the `request` and `reply` passed to this function. You have two options then: creating a function that returns another function, or defining the function inside the scope of this function, so it can access its parameters.

```javascript
// 1st option
function _doStuff(request, reply) {
  return function _foobar(_sources, _trials) {
    // ..
  }
}

function searchPage(request, reply) {
  return Promise.join(
    sources.list(),
    trials.search( /* ... */ ),
    _doStuff(request, reply)
  );
}

// 2nd option
function searchPage(request, reply) {
  function _fooBar(_sources, _trials) {
    // ...
  }

  return Promise.join(
    // ...
  );
}
```

I don't have a strong preference either way, or even on keeping the function as is.

handlers/search.js
nightsh: Oops, this slipped through. Will fix, thx

assets/js/index.js
nightsh: Alright, I'll check this and make sure you can't enter a non existent string.

handlers/search.js
nightsh: I had one reason to not do the extraction you mentioned: we don't normally do it in handlers, so it would be unusual throughout the handlers code. Everywhere else we launch a promise (if needed), followed by a `then` block where we call `reply.view`. It's just became a standard :)

vitorbaptista:
@nightsh Is this good to go? If so, could you please resolve the conflicts?

vitorbaptista:
Just missing tests :v:

vitorbaptista:
Fixed via https://github.com/opentrials/opentrials/commit/dbd05689d8c6e22ed7f88ea76100d276fd191544. 

I'm sorry, @nightsh. I didn't use GitHub's merge tool because I wanted to squash everything except the tests refactoring commit (https://github.com/opentrials/opentrials/commit/8cd82f36f732930e841adbfb30f0225e3c40348f). After I pushed to master I noticed the commit only has my name.

Issue #741, opened 2017-04-10
List our database dumps


vitorbaptista:
Since https://github.com/opentrials/opentrials/issues/488, we're saving regular DB dumps on S3. However, we don't list them anywhere.

# Tasks

- [x] Create "Data" page on the Explorer, containing the links to our current DB dumps
- [x] Write also how to load the data on Postgres. Keep it simple and link to Postgres' documentation as much as possible.

vitorbaptista:
Fixed (see https://explorer.opentrials.net/data)

PR #760, opened 2017-04-19
[#741] List our database dumps

nightsh:


nightsh:
We still need to brush up the text, but I clearly have no idea how to put it nicely, so I opened up the PR and waiting for suggestions :slightly_smiling_face: 
@vitorbaptista can you please have a look at it? Also, I couldn't figure out a way to provide a permalink to the dump, we might need to adjust the processor. Not sure that's the way, so asking here.

One more thing: I initially created this page mentioning warehouse and api databases, then I realised we don't dump warehouse. Is that temporary, or we won't dump warehouse anytime soon? (could be useful for people willing to try out deduplication and other data magic stuff on the raw-ish input warehouse is).

vitorbaptista:
@nightsh Ping me when you add the code to get the list of dumps from S3, then we can work on the actual text :v:

vitorbaptista:
@nightsh Could you take a look on https://github.com/opentrials/opentrials/pull/760/commits/fadf027976dee609c3ec83ddb3c2d90c16afe286? I ended up reorganizing your code:

* Move the S3 bucket's prefix from outside `listObjects()` code
* Move the logic to create the dump's URL and labels from the view to the handler
* Remove `#listObjects` test (couldn't mock it :/)
* Add tests to the handler (mocking S3, as otherwise we'll try to hit S3 servers when running the tests)
* Expand a bit on the text

handlers/data.js
```diff
@@ -0,0 +1,53 @@
...
+
+  s3.listObjects(DUMPS_PREFIX)
+    .catch((err) => {
+      // eslint-disable-next-line no-console
+      console.error('Could not get the list of database dumps:', err);
+      return {
+        Contents: [],
+      };
+    })
+    .then(_parseDumps)
```
nightsh: I like this idea, good practice, much win \o/

handlers/data.js
```diff
@@ -0,0 +1,53 @@
...
+      label: dump.Key.split('/').pop(),
+      size: dump.Size,
+    }
+  ));
+}
+
+module.exports = {
+  handler: data,
+  cache: {
+    expiresIn: 60 * 60 * 24 * 7,  // one week
```
nightsh: Is this actually working to cache the request (i.e. `dumps`) for a week or until the process is restarted? I never tried it, let alone testing it. If so, it's quite a nice touch!

handlers/data.js
```diff
@@ -0,0 +1,53 @@
...
+    privacy: 'private',
```
nightsh: What's this doing?

test/handlers/data.js
```diff
@@ -0,0 +1,72 @@
...
+const config = require('../../config');
+const s3 = require('../../agents/s3');
+
+
+describe('data handler', () => {
+  let server;
+  before(() => getExplorerServer().then((_server) => (server = _server)));
+
+  beforeEach(() => {
+    sinon.stub(s3, 'listObjects').resolves({
```
nightsh: So this is the much easier way to mock... will try to remember it, thx!

views/data.html
```diff
@@ -0,0 +1,76 @@
...
+      <p>
+        The entire OpenTrials database is saved monthly automatically. This
+        is what we use to power our API, and contains everything that you can
+        see on this website. Having the entire database in your local machine
+        allows you to analyze the data using your favorite tool, or even create
+        your own copy of the OpenTrials API.
+      </p>
+      <p>
+        As with any usage of OpenTrials's data, make sure you read and
+        understand our <a href="/terms-of-use">Terms of use</a>.
```
nightsh: Ugh, thanks, I forgot about this reference. Good thing you remembered.

views/data.html
```diff
@@ -0,0 +1,76 @@
...
+                {{ dump.label }}
+            </a> ({{ dump.size | formatBytes }})
+          </li>
+          {% endfor %}
+        </ul>
+      {% else %}
+        <p>
+          Unfortunately, something went wrong and the list of dumps could
+          not be obtained. Please try again later. If the issue remains,
+          please <a href="/flag-error?url={{ currentURL | urlencode }}">let us
```
nightsh: No github link?

agents/s3.js
```diff
@@ -87,7 +87,20 @@ function getSignedFormFields(additionalConditions) {
   };
 }
 
+function listObjects(prefix) {
+  const s3Client = new S3();
+  const params = {
+    Bucket: config.s3.bucket,
+    Delimiter: '/',
+    Prefix: prefix,
```
nightsh: Ah, nice catch, this way we can reuse this method even more, by changing the prefix. Silly me :-/

views/data.html
```diff
@@ -0,0 +1,76 @@
...
+    </section>
+    <section>
+      <h2>Download</h2>
+      {% if dumps.length > 0 %}
+        <p>
+          Below are the latest available dumps of our database. Their
+          file names contain the date at which the dump was produced:
+        </p>
+        <ul>
+          {% for dump in dumps | sort(true, false, 'label') %}
```
nightsh: An alternative way of displaying would be to generate a table and parse the date from the file names to use as a column. Easier on the eye to find the most or the least recent dump available. But if the number of dumps stays small, we're better off with the plain list..

handlers/data.js
vitorbaptista: This sets the cache headers of the HTTP request, so the user's browsers will cache it. It's limited for each user, though.

handlers/data.js
vitorbaptista: This sets the cache-control as private. It means that no intermediate proxies (e.g. CloudFlare) should cache this, only the user himself. This is because the page contain data that can be relevant for the user only. In our case, that's only the login user.

views/data.html
vitorbaptista: I think GitHub is too specific for techies, so it's better if people submit using our `flag-error` page.

Issue #717, opened 2017-04-03
Add search sort order at top of search results


benmeg:
There is a plan (#286) to allow ordering search results by other variables.

Until that is implemented, it would be helpful to indicate to users how the search results are ordered (trial registration date, descending). User testing has shown that for some users the order is not obvious.

Just a simple bit of text above the list of trials would suffice - for example:

_"Results ordered by trial registration date, descending"_

![screenshot_27](https://cloud.githubusercontent.com/assets/8491252/24621319/f36c3bac-1898-11e7-9626-517bb4128589.png)

PR #773, opened 2017-04-20
[#717] Add info about search results sort order

georgiana-b:
Fixes opentrials/opentrials#717

Simple writing was easy to miss so I styled it a bit:

search-info-box


nightsh:
I would go for a normal weight, italic text (paragraph) followed by a `<hr>`. Harder to miss, less colourful.

benmeg:
> I would go for a normal weight, italic text (paragraph) followed by a `<hr>`. Harder to miss, less colourful.

+1 on this

assets/styles/_search-results.scss
```diff
@@ -69,6 +69,12 @@
     margin: 0;
     list-style: none;
   }
+
+  .info {
```
vitorbaptista: The `info` name is too generic IMHO, so I would prefer something like `search-results-sort-order` or similar.

vitorbaptista:
Agreed with @nightsh.

georgiana-b:
How it looks after applying suggestions :point_up::

spectacle t16864

Issue #315, opened 2016-08-17
Add tooltips for each search filter


vitorbaptista:
They should briefly explain what the filter is about. For inspiration, you can check the filters on https://clinicaltrials.gov/ct2/search/advanced. For the design, you can dimply add a `?` symbol after the filter names, similar to how we did with the discrepancies:

![discrepancy icon design](https://cloud.githubusercontent.com/assets/76945/17748939/84637e1c-64b3-11e6-8f33-0447762f9af4.png)

smth:
@vitorbaptista now the design and code for this is done (and am not currently working on OpenTrials), it might be a good idea to assign to someone else?

vitorbaptista:
Thanks @smth! @opentrials/community Could you write a small text explaining each filter? We already have for "Conditions" and "Interventions" in http://explorer.opentrials.net/search, so need the rest (and those could be improved as well). The filters from ClinicalTrials.gov could serve as a starting point (see https://clinicaltrials.gov/ct2/search/advanced)

georgiana-b:
> @opentrials/community Could you write a small text explaining each filter? We already have for "Conditions" and "Interventions" in http://explorer.opentrials.net/search, so need the rest (and those could be improved as well)

@opentrials/community Do have these somewhere?

georgiana-b:
I'll move this to Blocked until we get the text for tooltips.

vitorbaptista:
Ping @opentrials/research.

There're two sides to this issue. One is explaining what the field is (e.g. "Conditions are diseases, disorders, syndromes, ..."), the other is explaining what it searches across (e.g. "Conditions searches across the conditions field of all trials that have a separate conditions field").

For this issue, we should describe what the field is, and can describe what it searches across in a separate issue (and possibly a separate page)

benmeg:
@vitorbaptista @georgiana-b Here are some suggestions: https://docs.google.com/document/d/1Vmc3qSFz5iXLcfQuKYHqFm5Ag5k7HStQoGbjRyIzqvQ/

vitorbaptista:
Also add a tooltip on the main search input box, the one that searches across all fields, as it's unclear that it does it.

PR #816, opened 2017-05-02
[#315] Added tooltips for search filters

benmeg:
Fixes opentrials/opentrials#315


views/partials/search-form-fields.html
```diff
@@ -96,7 +120,10 @@
   </div>
 
   <div class="filter">
-    <label for="has_published_results">Study results</label>
+    <label for="has_published_results"
+           class="hastip"
+           title="If the trial has posted results on ClinicalTrials.gov"
```
vitorbaptista: This isn't specific to CT.gov; this is if we have found the trial's results at all, wherever that comes from (could even be from a data donation).

vitorbaptista:
Thanks, @benmeg! Merged :beers:

PR #828, opened 2018-05-23
Updated consent manager and privacy policy 

roll:
- connects https://github.com/okfn/devops/issues/142


views/terms-of-use.html
```diff
@@ -23,25 +23,6 @@
   </em>
 </p>
 
-<section>
-  <h2>Privacy</h2>
```
roll: @pwalsh 
There is a dedicated terms of use (e.g. fd.io just links to the central one). 

Should we just preserve the existing one, remove the privacy section and add the central privacy policy link to the footer? (as it is in this PR)

pwalsh: @roll well spotted. Should also link terms of use back to central one.

